PR #1, opened 2019-03-06
Base interfaces and classes

arkivanov:


library/src/commonMain/kotlin/com/badoo/reaktive/disposable/Various.kt
```diff
@@ -0,0 +1,25 @@
+package com.badoo.reaktive.disposable
+
+import com.badoo.reaktive.utils.lock.newLock
+import com.badoo.reaktive.utils.lock.synchronized
+
+internal inline fun disposable(crossinline onDispose: () -> Unit = {}): Disposable =
+    object : Disposable {
+        private val lock = newLock()
+
+        private var isDisposed_ = false
```
lukaville: `isDisposed_` -> `_isDisposed`?
https://kotlinlang.org/docs/reference/properties.html

library/src/commonMain/kotlin/com/badoo/reaktive/disposable/DisposableWrapper.kt
```diff
@@ -0,0 +1,50 @@
...
+
+    private val lock = newLock()
+    private var _isDisposed = false
+    override val isDisposed: Boolean get() = lock.synchronized { _isDisposed }
+
+    override fun dispose() {
+        if (!_isDisposed) {
+            var d: Disposable? = null
+            lock.synchronized {
+                if (isDisposed) {
```
olegpetsjonkin: _isDisposed ?

library/src/commonMain/kotlin/com/badoo/reaktive/disposable/DisposableWrapper.kt
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Thread-safe container of one [Disposable]
+ */
+class DisposableWrapper(
+    private var disposable: Disposable? = null
+) : Disposable {
+
+    private val lock = newLock()
+    private var _isDisposed = false
```
mk-jones: Shouldn't this be @ Volatile?

library/src/commonMain/kotlin/com/badoo/reaktive/disposable/DisposableWrapper.kt
```diff
@@ -0,0 +1,50 @@
...
+                disposable = null
+            }
+            d?.dispose()
+        }
+    }
+
+    /**
+     * Atomically either sets the specified [Disposable] or disposes it if wrapper is already disposed
+     */
+    fun set(disposable: Disposable?): Disposable? {
```
mk-jones: I'd expect this to dispose the replaced observable for me.
```suggestion
    fun set(disposable: Disposable?) {
    if (!_isDisposed) {
            lock.synchronized {
                if (!_isDisposed) {
                    val old = this.disposable
                    this.disposable = disposable
                    old?.dispose()
                }
            }
        }
        ...
        }
```

library/src/commonMain/kotlin/com/badoo/reaktive/disposable/Various.kt
```diff
@@ -0,0 +1,26 @@
...
+
+import com.badoo.reaktive.utils.lock.newLock
+import com.badoo.reaktive.utils.lock.synchronized
+
+internal inline fun disposable(crossinline onDispose: () -> Unit = {}): Disposable =
+    object : Disposable {
+        private val lock = newLock()
+
+        @Suppress("ObjectPropertyName") // backing property
+        private var _isDisposed = false
```
mk-jones: Also needs volatile

library/src/commonMain/kotlin/com/badoo/reaktive/disposable/DisposableWrapper.kt
arkivanov: Don't think so, it's only accessed from synchronized block

library/src/commonMain/kotlin/com/badoo/reaktive/disposable/Various.kt
arkivanov: Discussed, not needed

library/src/jvmMain/kotlin/com/badoo/reaktive/scheduler/FixedExecutorServiceStrategy.kt
```diff
@@ -0,0 +1,26 @@
...
+            lazy { Executors.newSingleThreadScheduledExecutor(threadFactory) }
+        }
+
+    private val executorIndex = AtomicInteger()
+
+    override fun get(): ScheduledExecutorService =
+        pool[executorIndex.getAndIncrement() % pool.size].value
+
+    override fun recycle(executorService: ScheduledExecutorService) {
+        // no-op
```
mk-jones: What do you think of calling shutdown? For FixedThreadPools the docs state that the threads will exist until shutdown is called. Also ExecutorService has shutdown called in the finalize method. I think it would be a good idea to more reliably call it ourselves.

library/src/jvmMain/kotlin/com/badoo/reaktive/utils/ExpirationPool.kt
```diff
@@ -0,0 +1,94 @@
...
+        init {
+            isDaemon = true
+        }
+
+        override fun run() {
+            super.run()
+
+            while (!isInterrupted) {
+                try {
+                    queue.take()
```
mk-jones: Read above about shutdown.

PR #2, opened 2019-03-08
Observable MVP

arkivanov:


library/src/commonMain/kotlin/com/badoo/reaktive/observable/Sample.kt
```diff
@@ -0,0 +1,54 @@
...
+    observableByEmitter { emitter ->
+        val disposableWrapper = CompositeDisposable()
+        emitter.setDisposable(disposableWrapper)
+        val executor = scheduler.newExecutor()
+        disposableWrapper += executor
+
+        subscribeSafe(
+            object : ObservableObserver<T> {
+                private val lock = newLock()
+                private lateinit var lastValue: MutableList<T>
```
mk-jones: Why is this a list?

library/src/commonMain/kotlin/com/badoo/reaktive/observable/Throttle.kt
```diff
@@ -0,0 +1,20 @@
...
+import com.badoo.reaktive.utils.uptimeMillis
+
+fun <T> Observable<T>.throttle(window: Long): Observable<T> =
+    observable { observer ->
+        subscribeSafe(
+            object : ObservableObserver<T> by observer {
+                private var lastTime = 0L
+
+                override fun onNext(value: T) {
+                    val time = uptimeMillis
```
mk-jones: Would be easier for testing if we somehow avoided the singleton style of checking time.
If this could perhaps be passed inside an internal constructor or something.

library/src/commonMain/kotlin/com/badoo/reaktive/observable/Zip.kt
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+fun <T, R> zip(vararg sources: Observable<T>, mapper: (List<T>) -> R): Observable<R> =
+    sources.toList().zip(mapper)
+
+fun <T1, T2, R> zip(
+    source1: Observable<T1>,
+    source2: Observable<T2>,
+    mapper: (T1, T2) -> R
+): Observable<R> =
```
mk-jones: Could use reified to avoid unchecked cast.

library/src/commonMain/kotlin/com/badoo/reaktive/observable/Various.kt
```diff
@@ -0,0 +1,39 @@
+package com.badoo.reaktive.observable
+
+inline fun <T> observable(crossinline subscribe: (observer: ObservableObserver<T>) -> Unit): Observable<T> =
```
mk-jones: There is an extension function subscribe, a method subscribe and a variable function subscribe. It gets pretty confusing, I think it is worth renaming this variable. It currenty looks like the subscribe method recursively calls itself.

library/src/commonMain/kotlin/com/badoo/reaktive/observable/Zip.kt
```diff
@@ -0,0 +1,231 @@
...
+import com.badoo.reaktive.disposable.Disposable
+import com.badoo.reaktive.utils.lock.newLock
+import com.badoo.reaktive.utils.lock.synchronized
+
+fun <T, R> Collection<Observable<T>>.zip(mapper: (List<T>) -> R): Observable<R> =
+    observableByEmitter { emitter ->
+        val disposables = CompositeDisposable()
+        emitter.setDisposable(disposables)
+        val lock = newLock()
+        val values = List<MutableList<T>>(size) { mutableListOf() }
```
mk-jones: LinkedList instead of mutableListOf() as we always remove from the head.

library/src/commonMain/kotlin/com/badoo/reaktive/observable/Zip.kt
```diff
@@ -0,0 +1,231 @@
...
+                            values[index].add(value)
+
+                            if (values.all(List<*>::isNotEmpty)) {
+                                values.forEach { queue ->
+                                    readyValues.add(queue.removeAt(0))
+                                }
+
+                                mapper(readyValues)
+                                    .also {
+                                        readyValues.clear()
```
mk-jones: Consider a mapper that simply returns the combined list ( (list) -> R = list )
When we call clear we would be clearing the output leading to unexpected behaviour. Better to copy values to a new list before emitting.

library/src/commonMain/kotlin/com/badoo/reaktive/observable/CombineLatest.kt
```diff
@@ -0,0 +1,231 @@
...
+                    override fun onComplete() {
+                        lock.synchronized {
+                            if (--activeSourceCount == 0) {
+                                emitter.onComplete()
+                            }
+                        }
+                    }
+
+                    override fun onError(error: Throwable) {
+                        lock.synchronized {
```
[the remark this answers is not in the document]
arkivanov: 1. Lock is needed as there is a concurrency: multiple sources can emit errors simultaneously
2. As we are using emitter, it will dispose upstream on error. It is safe to emit after onError or onComplete, observableByEmiiter takes care of it.

library/src/commonMain/kotlin/com/badoo/reaktive/observable/Debounce.kt
```diff
@@ -0,0 +1,72 @@
+package com.badoo.reaktive.observable
+
+import com.badoo.reaktive.disposable.CompositeDisposable
+import com.badoo.reaktive.disposable.Disposable
+import com.badoo.reaktive.scheduler.Scheduler
+import com.badoo.reaktive.utils.lock.newLock
+import com.badoo.reaktive.utils.lock.synchronized
+
+fun <T> Observable<T>.debounce(timeout: Long, scheduler: Scheduler): Observable<T> =
```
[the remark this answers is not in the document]
arkivanov: I found that most of the people does not set scheduler and at the same time expect values to be emitted on previous one (e.g. main thread). I wanted to make all schedulers explicit to improve error-prone. What do you think?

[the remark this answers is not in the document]
arkivanov: Will rename to timeoutMillis, we don't have TimeUnit

library/src/commonMain/kotlin/com/badoo/reaktive/observable/FlatMap.kt
```diff
@@ -0,0 +1,60 @@
...
+                    try {
+                        mapper(value)
+                    } catch (e: Throwable) {
+                        onError(e)
+                        return
+                    }
+                        .subscribeSafe(
+                            object : ObservableObserver<R>, CompletableObserver by this {
+                                override fun onNext(value: R) {
+                                    lock.synchronized {
```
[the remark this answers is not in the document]
arkivanov: Callbacks of emitter must not be called concurrently (more than one callback at a time), only synchronously. Here we have multiple sources that can emit at any time, we must synchronise them.

library/src/commonMain/kotlin/com/badoo/reaktive/observable/FlatMap.kt
```diff
@@ -0,0 +1,60 @@
...
+                override fun onComplete() {
+                    lock.synchronized {
+                        if (--activeSourceCount == 0) {
+                            emitter.onComplete()
+                        }
+                    }
+                }
+
+                override fun onError(error: Throwable) {
+                    lock.synchronized {
```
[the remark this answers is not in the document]
arkivanov: Checkout my comment above

library/src/commonMain/kotlin/com/badoo/reaktive/observable/ObserveOn.kt
```diff
@@ -0,0 +1,40 @@
...
+fun <T> Observable<T>.observeOn(scheduler: Scheduler): Observable<T> =
+    observableByEmitter { emitter ->
+        val disposables = CompositeDisposable()
+        emitter.setDisposable(disposables)
+        val executor = scheduler.newExecutor()
+        disposables += executor
+
+        subscribeSafe(
+            object : ObservableObserver<T> {
+                private val bufferedExecutor = BufferedExecutor(executor, emitter::onNext)
```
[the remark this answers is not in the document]
arkivanov: This is the point of Observable (same in RxJava), there is not backpressure control. We can introduce Flowable later on which is backpressure aware.

library/src/commonMain/kotlin/com/badoo/reaktive/observable/Sample.kt
arkivanov: We need some kind of mutable holder of nullable value so we can distinguish between the following cases:
1. There is no value at all (not initialised)
2. There is a value, null or not null.

library/src/commonMain/kotlin/com/badoo/reaktive/observable/Throttle.kt
arkivanov: I thought about it, but there is no constructor here at all. We probably can mock "static" time provider in tests. Or we can do same with RxJava2: use scheduler to open gate again when timeout passed. Then we can actually remove uptimeMillis at all as this is the only usage at the moment.

library/src/commonMain/kotlin/com/badoo/reaktive/observable/Zip.kt
arkivanov: There is no LinkedList nor any queue in Kotlin native, I will try to write simple queue

library/src/commonMain/kotlin/com/badoo/reaktive/observable/Zip.kt
arkivanov: I did not get this, could you explain?

library/src/commonMain/kotlin/com/badoo/reaktive/observable/CombineLatest.kt
```diff
@@ -0,0 +1,231 @@
...
+                values[3] as T4,
+                values[4] as T5,
+                values[5] as T6,
+                values[6] as T7,
+                values[7] as T8,
+                values[8] as T9
+            )
+        }
+
+fun <T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, R> combineLatest(
```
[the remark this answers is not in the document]
arkivanov: It will require making function inline which will produce incorrect stack traces if mapper throws an exception. Casting is actually safe in these cases.

library/src/commonMain/kotlin/com/badoo/reaktive/observable/Zip.kt
arkivanov: See my comment for ConbineLatest

library/src/commonMain/kotlin/com/badoo/reaktive/observable/CombineLatest.kt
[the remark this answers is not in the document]
arkivanov: Also there is a limitation, the following code won't compile:
```
fun <T> foo(v: T) {
    bar(v)
}

inline fun <reified T> bar(v: T) {
}
```

library/src/commonMain/kotlin/com/badoo/reaktive/observable/Zip.kt
arkivanov: I've got it

library/src/commonMain/kotlin/com/badoo/reaktive/observable/Throttle.kt
mk-jones: Makes sense to remove the uptimeMillis.

library/src/commonMain/kotlin/com/badoo/reaktive/observable/Zip.kt
mk-jones: Ah, hadn't thought of that. Perhaps we could use an expect or something?

library/src/commonMain/kotlin/com/badoo/reaktive/observable/Throttle.kt
arkivanov: I would like to discuss this one, we can fix it separately later on.

arkivanov:
Unresolved questions are to be discussed. Any fixes will be in separate PRs.

PR #3, opened 2019-03-13
Android main scheduler and sample app

arkivanov:


library/src/androidMain/kotlin/com/badoo/reaktive/scheduler/MainScheduler.kt
```diff
@@ -0,0 +1,75 @@
...
+    override fun newExecutor(): Scheduler.Executor =
+        ExecutorImpl()
+            .also(disposables::add)
+
+    override fun destroy() {
+        disposables.dispose()
+    }
+
+    private class ExecutorImpl : Scheduler.Executor {
+        @Volatile
```
olegpetsjonkin: Do we need Volatile here?

arkivanov: Absolutely, as it's accessed from isDisposed's get method where there is no synchronisation.

library/src/androidMain/kotlin/com/badoo/reaktive/scheduler/MainScheduler.kt
```diff
@@ -0,0 +1,75 @@
...
+
+        override fun cancel() {
+            executeIfNotRecycled {
+                it.removeCallbacksAndMessages(null)
+            }
+        }
+
+        private inline fun <T> executeIfNotRecycled(block: (Handler) -> T): T? {
+            if (handler != null) {
+                synchronized(monitor) {
```
mk-jones: Why synch instead of a lock?

arkivanov: Because we have proper synchronisation in JVM

PR #6, opened 2019-03-21
Completable MVP

arkivanov:


library/src/commonMain/kotlin/com/badoo/reaktive/completable/AsObservable.kt
```diff
@@ -0,0 +1,9 @@
+package com.badoo.reaktive.completable
+
+import com.badoo.reaktive.observable.Observable
+import com.badoo.reaktive.observable.observable
+
+fun <T> Completable.asObservable(): Observable<T> =
```
mk-jones: The comparable RxJava methods would be toObservable() right? I think we should keep the same naming scheme.

arkivanov: We have a method clash between toObservable and asObservable so we need to distinguish somehow...

PR #7, opened 2019-03-22
Additional operators for MVP

arkivanov:


library/src/commonMain/kotlin/com/badoo/reaktive/utils/SortOrder.kt
```diff
@@ -0,0 +1,5 @@
+package com.badoo.reaktive.utils
+
+enum class SortOrder {
```
lukaville: unused?

arkivanov: Yep, will remove

arkivanov: Fixed

library/src/commonMain/kotlin/com/badoo/reaktive/completable/DoOnBefore.kt
```diff
@@ -0,0 +1,109 @@
...
+                }
+            }
+        )
+    }
+
+fun Completable.doOnBeforeFinally(action: () -> Unit): Completable =
+    completable { observer ->
+        subscribeSafe(
+            object : CompletableObserver by observer {
+                private var lock = newLock()
```
mk-jones: val

library/src/commonMain/kotlin/com/badoo/reaktive/completable/DoOnBefore.kt
```diff
@@ -0,0 +1,109 @@
...
+        subscribeSafe(
+            object : CompletableObserver by observer {
+                override fun onSubscribe(disposable: Disposable) {
+                    observer.onSubscribe(disposable.wrap(onBeforeDispose = action))
+                }
+            }
+        )
+    }
+
+fun Completable.doOnBeforeFinally(action: () -> Unit): Completable =
```
mk-jones: Could you me the difference between this one and beforeTerminate? It sort of seems like they should a single method.

library/src/commonMain/kotlin/com/badoo/reaktive/disposable/DisposableExt.kt
```diff
@@ -0,0 +1,13 @@
+package com.badoo.reaktive.disposable
+
+internal inline fun Disposable.wrap(
+    crossinline onBeforeDispose: () -> Unit = {},
```
mk-jones: nullable instead of empty function?

library/src/commonMain/kotlin/com/badoo/reaktive/observable/DoOnBefore.kt
```diff
@@ -0,0 +1,121 @@
...
+            object : ObservableObserver<T> by observer {
+                override fun onSubscribe(disposable: Disposable) {
+                    action(disposable)
+                    observer.onSubscribe(disposable)
+                }
+            }
+        )
+    }
+
+fun <T> Observable<T>.doOnBeforeNext(consumer: (T) -> Unit): Observable<T> =
```
mk-jones: Couldn't this just be doOnNext? Would the user ever care about the "before" part on most of these methods.

library/src/commonMain/kotlin/com/badoo/reaktive/completable/DoOnBefore.kt
arkivanov: Same with RxJava, terminate is when stream reached the end (onComplete, onError), finally is when stream finished (either reached the end or disposed)

library/src/commonMain/kotlin/com/badoo/reaktive/disposable/DisposableExt.kt
arkivanov: Not possible with inline

library/src/commonMain/kotlin/com/badoo/reaktive/observable/DoOnBefore.kt
arkivanov: I wanted to stress that it's exactly "before". Later may add events like onAfterTerminate (like in RxJava).

library/src/commonMain/kotlin/com/badoo/reaktive/completable/DoOnBefore.kt
arkivanov: Fixed

PR #8, opened 2019-03-22
RxJava interop

arkivanov:


rxjava-interop/src/main/kotlin/com/badoo/reaktive/rxjavainterop/Completable.kt
```diff
@@ -0,0 +1,53 @@
+package com.badoo.reaktive.rxjavainterop
```
lukaville: probably we should also add interop with rxjava schedulers to reuse the thread pools

arkivanov: Can you tell any use cases?

lukaville: currently, we reuse rxjava 1/2 schedulers so we don't have duplicated thread pools

arkivanov: Sounds reasonable, would it be enough to just convert rxjava2 schedulers to Reaktive. So Reaktive could reuse rxjava2's schedules.

arkivanov: I have pushed the changes, could you please have a look?

PR #9, opened 2019-03-26
Proper serialization of concurrent sources

arkivanov:


library/src/commonMain/kotlin/com/badoo/reaktive/observable/CombineLatest.kt
```diff
@@ -2,19 +2,62 @@ package com.badoo.reaktive.observable
...
+                            }
+                            ?.also(emitter::onNext)
+
+                        true
+                    }
+
+                    is CombineLatestEvent.OnComplete -> {
+                        activeSourceCount--
+
+                        val allCompleted = (activeSourceCount == 0) || (values[event.index] === dummyValue)
```
mk-jones: (values[event.index] === dummyValue)
From what I understand if it hasn't emitted a value and it completes empty, then we must complete the entire thing. 
Not obvious, worth adding a quick comment.

library/src/commonMain/kotlin/com/badoo/reaktive/observable/Zip.kt
```diff
@@ -6,16 +6,65 @@ import com.badoo.reaktive.utils.arrayqueue.ArrayQueue
...
+                        }
+
+                        true
+                    }
+
+                    is ZipEvent.OnComplete -> {
+                        completed[event.index] = true
+
+                        val allCompleted = completed.all { it }
+                        if (allCompleted) {
```
mk-jones: allCompleted || values[event.index].isEmpty()

library/src/commonMain/kotlin/com/badoo/reaktive/utils/serializer/Serializer.kt
```diff
@@ -0,0 +1,82 @@
...
+    /**
+     * Called synchronously for every value
+     *
+     * @param value a value
+     * @return true if processing should continue, false otherwise
+     */
+    protected abstract fun onValue(value: T): Boolean
+
+    private companion object {
+        private inline fun <T> ArrayQueue<T>.drain(
```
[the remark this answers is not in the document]
arkivanov: I believe it's a good style to put functions that does not use anything from main object to companions. It makes reading a bit easier as you understand that fact. Plus it's a bit more error-pron.

PR #11, opened 2019-03-29
Subjects MVP

arkivanov:


library/src/commonMain/kotlin/com/badoo/reaktive/subject/DefaultSubject.kt
```diff
@@ -0,0 +1,128 @@
...
+import com.badoo.reaktive.disposable.DisposableWrapper
+import com.badoo.reaktive.disposable.disposable
+import com.badoo.reaktive.observable.ObservableObserver
+import com.badoo.reaktive.utils.lock.Lock
+import com.badoo.reaktive.utils.lock.newLock
+import com.badoo.reaktive.utils.lock.synchronized
+import com.badoo.reaktive.utils.serializer.serializer
+import com.badoo.reaktive.utils.synchronizedReadWriteProperty
+
+internal open class DefaultSubject<T> : Subject<T> {
```
lukaville: Let's add Relay out of the box? Or do you have plans to do it in the following PRs?

arkivanov: I don't see any points of relay in our implementation but we can discuss it and do in a separate PR.

PR #12, opened 2019-03-29
Added Trampoline scheduler for JVM

arkivanov:


library/src/commonMain/kotlin/com/badoo/reaktive/utils/queue/ArrayQueue.kt
```diff
@@ -1,18 +1,19 @@
-package com.badoo.reaktive.utils.arrayqueue
+package com.badoo.reaktive.utils.queue
 
 import kotlin.math.abs
 
-internal class ArrayQueue<T> {
+// FIXME: internal
```
lukaville: internal?

PR #18, opened 2019-04-11
Travis CI support and not implemented iOS target

CherryPerry:
iOS targets are only built, not tested. When tests are ready, see how to launch tests [here](https://github.com/russhwolf/multiplatform-settings/blob/master/multiplatform-settings/build.gradle#L96).

Add tag to release after commit check is finished, otherwise only part of modules can be released.

Inline/Crossline compilation error on native target is discovered, should be reported to Kotlin before merge or after (make decision please).

Now lib is published to local repository on CI to check if it even works (it does). Bintray should be implemented later when registration is completed.


target.gradle
```diff
@@ -0,0 +1,30 @@
...
+
+    boolean isIosSim() { return this == ALL || this == IOS_SIM }
+
+    boolean isMeta() { return this == META }
+}
+
+ext.currentTarget = { ->
+    final value = System.getProperty('MP_TARGET')
+    final target = MpBuildTarget.values().find { it.name().equalsIgnoreCase(value) }
+    if (target != null) {
```
lukaville: http://groovy-lang.org/operators.html#_elvis_operator

CherryPerry:
@arkivanov Can be reviewed and merged now.

Tests from commonTest are launched on all platforms now except android emulator.
Inline/crossinline compilation error is fixed in 1.3.30.

reaktive/src/commonTest/kotlin/com/badoo/reaktive/utils/arrayqueue/ArrayQueueTest.kt
```diff
@@ -10,24 +11,28 @@ class ArrayQueueTest {
     private val queue = ArrayQueue<String?>()
 
     @Test
+    @JsName("initialSizeIsZero")
```
lukaville: we'll have the same problem with android tests. maybe let's use camel case naming by default?

CherryPerry: Good idea

publication.gradle
```diff
@@ -0,0 +1,29 @@
+apply from: '../target.gradle'
+final target = currentTarget()
+
+kotlin {
+    final taskConfigurationMap = [
+            'kotlinMultiplatform': target.isMeta(),
+            'metadata'           : target.isMeta(),
+            'jvm'                : target.isJvm(),
+            'js'                 : target.isJvm(),
```
arkivanov: Why isJvm for JS? Looks like we need isJs, or just rename it to something common.

reaktive/src/iosCommonMain/kotlin/com/badoo/reaktive/scheduler/Schedulers.kt
```diff
@@ -0,0 +1,29 @@
+package com.badoo.reaktive.scheduler
+
+/**
+ * Creates a new instance of Main [Scheduler]
+ */
+actual fun createMainScheduler(): Scheduler {
```
arkivanov: It's not a good idea to create classes with same names in same packages. E.g. for JVM compilation fails with message "Duplicate JVM class name". Let's put each function into a separate file (like CreateComputationScheduler.kt, CreateIoScheduler.kt, etc.).

reaktive/build.gradle
```diff
@@ -1,72 +1,127 @@
...
             dependencies {
                 implementation 'org.jetbrains.kotlin:kotlin-stdlib-common'
             }
         }
-
         commonTest {
             dependencies {
                 implementation 'org.jetbrains.kotlin:kotlin-test'
-                implementation "org.jetbrains.kotlin:kotlin-test-junit"
+                implementation 'org.jetbrains.kotlin:kotlin-test-junit'
```
arkivanov: :+1:

PR #20, opened 2019-04-16
Rename some factory methods

arkivanov:
It would be nice to rename some of our source factory methods so they will be more like in RxJava. E.g.:
`observableByEmitter{}` -> `observable{}`
`observable{}` -> `observableUnsafe{}`
etc.


mk-jones:
It would be helpful if commit messages + pull request would also briefly explain why changes are being made.

PR #21, opened 2019-04-17
Fixed bugs in ArrayQueue

arkivanov:
There are two edge case issues with ArrayQueue.

**Bug 1**
STR:
1. Offer 7 items to ArrayQueue
2. Poll 7 items
3. Offer 1 item

AR: ArrayQueue.size returns 7
ER: ArrayQueue.size returns 1

**Bug 2**
STR: Offer 8 items to ArrayQueue
AR: ArrayQueue.size returns 0
ER: ArrayQueue.size returns 8

### DIFF
--- a/reaktive/src/commonMain/kotlin/com/badoo/reaktive/utils/queue/ArrayQueue.kt
+++ b/reaktive/src/commonMain/kotlin/com/badoo/reaktive/utils/queue/ArrayQueue.kt
@@ -6,9 +6,15 @@ internal class ArrayQueue<T> : Queue<T> {
     private var head = 0
     private var tail = 0
     private var isFull = false
+    override val peek: T? get() = queue[head]
 
-    val peek: T? get() = queue[head]
-    override val size: Int get() = if (isFull) queue.size else ((tail - head) and (queue.size - 1))
+    override val size: Int
+        get() =
+            when {
+                isFull -> queue.size
+                tail >= head -> tail - head
+                else -> queue.size + tail - head
+            }
 
     override fun offer(item: T) {
         ensureCapacity()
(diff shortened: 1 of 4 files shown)


reaktive/src/commonMain/kotlin/com/badoo/reaktive/utils/queue/ArrayQueue.kt
```diff
@@ -1,16 +1,14 @@
...
 
-    private var queue: Array<T?> = createArray(8)
+    private var queue: Array<T?> = createArray(INITIAL_CAPACITY)
     private var head = 0
     private var tail = 0
     private var isFull = false
 
     val peek: T? get() = queue[head]
-    override val size: Int get() = abs(tail - head)
+    override val size: Int get() = if (isFull) queue.size else ((tail - head) and (queue.size - 1))
```
mk-jones: What was the bug and how does this fix it? Would be useful if in issue or commit the bug was mentioned.

arkivanov: Updated description of this issue, thanks

mk-jones: Wouldn't it be easier to track the size in a variable? Is there an advantage to this approach?
Also a comment would be good for ((tail - head) and (queue.size - 1)). It's a pretty cool trick, took me a while.

arkivanov: For me it looks more safe when you calculate instead of track. You can easily getting an inconsistency. I would like to even remove isFull flag, but it appears to be impossible.

arkivanov: Fixed in a bit different way, please check

arkivanov:
@mk-jones could you please review the changes?

mk-jones:
The fix looks ok. I think we should consider revisiting this class in the future - part of why the bug was missed was the not obvious way it is implemented.

PR #23, opened 2019-04-17
Fixed ExecutorServiceScheduler Future isDisposed always false

### LINKED ISSUE #22: Disposable future isDisposed will be always false
https://github.com/badoo/Reaktive/blob/bcba8b9ceeeb026d63f1304524234cba8abead17/reaktive/src/jvmCommonMain/kotlin/com/badoo/reaktive/scheduler/ExecutorServiceScheduler.kt#L81

Should it be like that?

```kotlin
override val isDisposed: Boolean get() = isDone
```

arkivanov:
Fixes #22

### DIFF
--- a/reaktive/src/jvmCommonMain/kotlin/com/badoo/reaktive/scheduler/ExecutorServiceScheduler.kt
+++ b/reaktive/src/jvmCommonMain/kotlin/com/badoo/reaktive/scheduler/ExecutorServiceScheduler.kt
@@ -78,7 +78,7 @@ internal class ExecutorServiceScheduler(
         private companion object {
             private fun Future<*>.toDisposable(): Disposable =
                 object : Disposable {
-                    override val isDisposed: Boolean = isDone
+                    override val isDisposed: Boolean get() = isDone
 
                     override fun dispose() {
                         cancel(true)


reaktive/src/jvmCommonMain/kotlin/com/badoo/reaktive/scheduler/ExecutorServiceScheduler.kt
```diff
@@ -78,7 +78,7 @@ internal class ExecutorServiceScheduler(
         private companion object {
             private fun Future<*>.toDisposable(): Disposable =
                 object : Disposable {
-                    override val isDisposed: Boolean = isDone
+                    override val isDisposed: Boolean get() = isDone
```
lukaville: can we cover this change with a test?

arkivanov: We'll cover everything with tests soon, do you really want to cover this particular case?

lukaville: okay, let's cover it later
okay

PR #25, opened 2019-04-24
Tests for ObservableByEmitter

arkivanov:
- Added MockK test dependency for common tests
- Added tests for ObservableByEmitter
- Renamed test functions in ArrayQueueTest for consistency

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -27,6 +27,7 @@ allprojects {
             force 'com.android.support.constraint:constraint-layout:1.1.3'
             force 'io.reactivex.rxjava2:rxjava:2.2.7'
             force 'io.reactivex.rxjava2:rxandroid:2.1.1'
+            force 'io.mockk:mockk:1.9.3.kotlin12'
         }
     }
 }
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -27,6 +27,7 @@ allprojects {
             force 'com.android.support.constraint:constraint-layout:1.1.3'
             force 'io.reactivex.rxjava2:rxjava:2.2.7'
             force 'io.reactivex.rxjava2:rxandroid:2.1.1'
+            force 'io.mockk:mockk:1.9.3.kotlin12'
```
CherryPerry: We use Kotlin 1.30, not 1.20, so use 1.9.3 version without suffix.
For multiplatform use "io.mockk:mockk-common:{version}".
Does it work on native platform? Does't seem so, as I can see only jvm/js is supported. Probably we need to use stubs.

reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/ObservableByEmitterTest.kt
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    @Test
+    fun emitted_same_values_and_completed_in_the_same_order() {
+        emitter.onNext(1)
+        emitter.onNext(2)
+        emitter.onNext(3)
+        emitter.onComplete()
+
+        verify(ordering = Ordering.ORDERED) {
```
lukaville: let's create TestObserver similar to RxJava TestObserver http://reactivex.io/RxJava/javadoc/io/reactivex/observers/TestObserver.html

PR #27, opened 2019-04-27
Configuration for jitpack and iOS target is enabled explicitly only

CherryPerry:
Now jitpack will build only reaktive and RxJava2 interoperability lib without launching tests, it's CI's job.
iOS support now can be enabled only explicitly to avoid iOS metadata publication.


arkivanov:
@lukaville Please have a look this one as well

PR #28, opened 2019-04-27
Upgrade Gradle version to 5.4

CherryPerry:
Looks like it works absolutely fine.


arkivanov:
@lukaville Please have a look

PR #32, opened 2019-05-03
Follow up to deliver subscribe errors to downstreams

arkivanov:
In this PR https://github.com/badoo/Reaktive/pull/29 I forgot about Single. Here I'm fixing Single as well.


build.gradle
```diff
@@ -29,4 +29,18 @@ allprojects {
             force 'io.reactivex.rxjava2:rxandroid:2.1.1'
         }
     }
+
+    // TODO remove once https://github.com/Kotlin/kotlin-frontend-plugin/issues/141 is fixed
+    plugins.whenPluginAdded { plugin ->
```
arkivanov: Currently sample-js-browser-app does not compile. This is the fix.

PR #34, opened 2019-05-05
Immutability + native + linuxX64 and linuxArm32Hfp targets

arkivanov:
- converted all the code to immutable
- added native common implementations
- added linuxX64 and linuxArm32Hfp targets
- added Single.blockingGet() operator
- added Observable.withLatestFrom() operator
- added TestObservable


CherryPerry:
Native common compilation failed for iOS.
> e: /Users/travis/build/badoo/Reaktive/reaktive/src/nativeCommonMain/kotlin/com/badoo/reaktive/utils/Condition.kt: (31, 25): Type mismatch: inferred type is Long but Int was expected

CherryPerry:
Or we can just drop x32 iOS support as not actual, latest iOS version does not supports it anyway.

arkivanov:
> Native common compilation failed for iOS.
> 
> > e: /Users/travis/build/badoo/Reaktive/reaktive/src/nativeCommonMain/kotlin/com/badoo/reaktive/utils/Condition.kt: (31, 25): Type mismatch: inferred type is Long but Int was expected
> 
> Try to build linuxX32.

There is no Linux x86 support, but I did try Linux ARM 32 and it fails with the same error. I could not find a solution for that. The reason is: types `__syscall_slong_t` and `__time_t` are defined as Long on x64 and as Int on x32. Perhaps we should drop iOS x32.

CherryPerry:
@arkivanov 

Then remove `reaktive:ios32MainBinaries` from `.travis.yml` and `ios32` target from `build.gradle` and `publication.gradle`.

arkivanov:
> @arkivanov
> 
> Then remove `reaktive:ios32MainBinaries` from `.travis.yml` and `ios32` target from `build.gradle` and `publication.gradle`.

I have solved the problem and added `linuxArm32Hfp` target :-) No need to drop x32 at the moment.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/Throttle.kt
```diff
@@ -11,16 +12,16 @@ fun <T> Observable<T>.throttle(windowMillis: Long): Observable<T> =
...
 
                 override fun onSubscribe(disposable: Disposable) {
                     disposableWrapper.set(disposable)
                 }
 
                 override fun onNext(value: T) {
                     val time = uptimeMillis
-                    if (time - lastTime >= windowMillis) {
-                        lastTime = time
+                    if (time - lastTime.value >= windowMillis) {
```
CherryPerry: Race condition. Probably can be fixed by:
```kotlin
...
val lastTimeValue = lastTime.value
if (time - lastTimeValue >= windowMillis) {
  lastTime.compareAndSet(lastTimeValue, time) {
    observer.onNext(value)
  }
}

```

reaktive/src/commonMain/kotlin/com/badoo/reaktive/single/BlockingGet.kt
```diff
@@ -0,0 +1,58 @@
+package com.badoo.reaktive.single
+
+import com.badoo.reaktive.disposable.Disposable
+import com.badoo.reaktive.utils.Lock
+import com.badoo.reaktive.utils.atomicreference.AtomicReference
+import com.badoo.reaktive.utils.synchronized
+
+fun <T> Single<T>.blockingGet(): T {
+    val lock = Lock()
```
CherryPerry: If we still have locks, why you don't use same the same way in was before in some dangerous for race condition places?

reaktive/src/jvmCommonMain/kotlin/com/badoo/reaktive/utils/PrioritySerializer.kt
```diff
@@ -0,0 +1,59 @@
+package com.badoo.reaktive.utils
+
+import java.util.*
```
CherryPerry: Don't use `*` imports. Disable this setting in IDEA Kotlin Code Style settings.

reaktive/src/nativeCommonMain/kotlin/com/badoo/reaktive/scheduler/CreateTrampolineScheduler.kt
```diff
@@ -0,0 +1,3 @@
+package com.badoo.reaktive.scheduler
+
+actual fun createTrampolineScheduler(): Scheduler = TODO("No trampoline scheduler so far :-(")
```
CherryPerry: Can we resuse `MainScheduler` with warning output in console? As I remember, we decided to not crash app, if something can't be implemented on platform. Otherwise client code will be covered with a lot of checks `if (platform) oneScheduler else otherScheduler`.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/completable/DoOnBefore.kt
```diff
@@ -100,22 +99,13 @@ fun Completable.doOnBeforeDispose(action: () -> Unit): Completable =
 
 fun Completable.doOnBeforeFinally(action: () -> Unit): Completable =
     completableUnsafe { observer ->
-        val lock = newLock()
-        var isFinished = false
+        val isFinished = AtomicReference(false)
```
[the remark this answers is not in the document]
arkivanov: Thought about it, do think there are much benefits. On JVM Boolean is cached. On Native - not sure, comparing to freezing and immutability - it's nothing.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/Collect.kt
```diff
@@ -2,27 +2,33 @@ package com.badoo.reaktive.observable
...
+                private val collection = AtomicReference(initialCollection, true)
+
                 override fun onSubscribe(disposable: Disposable) {
-                    disposableWrapper.set(disposable)
+                    emitter.setDisposable(disposable)
                 }
 
                 override fun onNext(value: T) {
-                    accumulator(collection, value)
+                    collection.value =
```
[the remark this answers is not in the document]
arkivanov: `onNext` must not be invoked simultaneously by Rx contract. All methods of `Observer` and `Emitter` must be called sequentially.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/Throttle.kt
arkivanov: See comment above, there is only one upstream

reaktive/src/commonMain/kotlin/com/badoo/reaktive/single/BlockingGet.kt
arkivanov: See comments above. Here we need `Lock` as it how `Condition` works.

reaktive/src/jvmCommonMain/kotlin/com/badoo/reaktive/utils/PrioritySerializer.kt
arkivanov: Fixed. Perhaps we should commit IDEA settings to the repo.

reaktive/src/nativeCommonMain/kotlin/com/badoo/reaktive/scheduler/CreateTrampolineScheduler.kt
arkivanov: I believe we should just implement it :-) Main scheduler violates Trampoline scheduler rules (it should execute in the same callstack)

arkivanov: But yes, better to use MainScheduler

arkivanov: Not sure how to produce a warning, let's keep it like this and then just implement it? It's quite easy.

CherryPerry: https://github.com/badoo/Reaktive/issues/41 Created issue, so we won't forget about it before release. I think it is good idea to create issues for each TODO and goals to make roadmap and process visible.

reaktive/src/jvmCommonMain/kotlin/com/badoo/reaktive/utils/PrioritySerializer.kt
CherryPerry: And then have problems with `.codestyle` file on different versions of IDEA and Android Studio?)

PR #38, opened 2019-05-06
Scan operator

Lamartio:


reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/Scan.kt
```diff
@@ -0,0 +1,41 @@
+package com.badoo.reaktive.observable
+
+import com.badoo.reaktive.completable.CompletableCallbacks
+import com.badoo.reaktive.disposable.Disposable
+import com.badoo.reaktive.utils.Uninitialized
+import com.badoo.reaktive.utils.lock.newLock
+import com.badoo.reaktive.utils.lock.synchronized
+
+@Suppress("UNCHECKED_CAST")
```
arkivanov: Please apply suppress annotations to statements, it will help preventing unsafe casts in other places.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/Scan.kt
```diff
@@ -0,0 +1,41 @@
...
+    observable { emitter ->
+        subscribeSafe(
+            object : ObservableObserver<T>, CompletableCallbacks by emitter {
+                val lock = newLock()
+                var cache: Any? = Uninitialized
+
+                override fun onNext(value: T) {
+                    val previous = cache
+                    val next = try {
+                        when (previous) {
```
arkivanov: Let's use `if-else` for such a simple `when`

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/Scan.kt
```diff
@@ -0,0 +1,41 @@
...
+                            Uninitialized -> value
+                            else -> accumulate(previous as T, value)
+                        }
+                    } catch (e: Throwable) {
```
arkivanov: After you change it to `if-else` then you can apply `try-catch` only to `accumulate` call.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/Scan.kt
```diff
@@ -0,0 +1,41 @@
...
+@Suppress("UNCHECKED_CAST")
+fun <T> Observable<T>.scan(accumulate: (acc: T, value: T) -> T): Observable<T> =
+    observable { emitter ->
+        subscribeSafe(
+            object : ObservableObserver<T>, CompletableCallbacks by emitter {
+                val lock = newLock()
+                var cache: Any? = Uninitialized
+
+                override fun onNext(value: T) {
+                    val previous = cache
```
arkivanov: Race condition: you are reading `cache` without `lock`. Writing is fine.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/ScanWithSeed.kt
```diff
@@ -0,0 +1,40 @@
+package com.badoo.reaktive.observable
+
+import com.badoo.reaktive.completable.CompletableCallbacks
+import com.badoo.reaktive.disposable.Disposable
+import com.badoo.reaktive.utils.lock.newLock
+import com.badoo.reaktive.utils.lock.synchronized
+
+fun <T, R> Observable<T>.scan(seed: R, accumulator: (acc: R, value: T) -> R): Observable<R> =
```
arkivanov: Let's put all `scan` operators into into one `Scan.kt` file, it's not so big.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/ScanWithSeed.kt
```diff
@@ -0,0 +1,40 @@
...
+    scan<T, R>({ seed }, accumulator)
+
+fun <T, R> Observable<T>.scan(getSeed: () -> R, accumulate: (acc: R, value: T) -> R): Observable<R> =
```
arkivanov: Looks like we can avoid some boilerplate code here. Could you please try making `getSeed` lambda optional (nullable) and just call this method from the very first `scan` operator (which is without any `seed`)?

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/ScanWithSeed.kt
```diff
@@ -0,0 +1,40 @@
...
+                    lock.synchronized {
+                        cache = next
+                    }
+
+                    emitter.onNext(next)
+                }
+
+                override fun onSubscribe(disposable: Disposable) {
+                    emitter.setDisposable(disposable)
+                    emitter.onNext(cache)
```
arkivanov: Better to move this call right before `subscribeSafe` call. It's ok to take initial seed at that point and emit it to downstream, as you are using safe (`emitter` based) version of `observable` factory.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/ScanWithSeed.kt
Lamartio: I agree that it seems that there is some code duplication. The problem is that the upstream is of type `T` and down is `R`. If I drop the seed, then there is no way to transform `T` to `R`.

arkivanov: Agree

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/ScanWithSeed.kt
Lamartio: True!

PR #36, opened 2019-05-05
Tests for ObservableByEmitter

arkivanov:
Added `TestObservableObserver` and a number of extension methods for it. Wrote tests for `observable { emitter }` factory method.

PR #39, opened 2019-05-06
Add withLatestFrom operator and basic tests

Lamartio:


arkivanov:
Thanks for your contributions! We'll review your PRs whenever we have time. Our goals at the moment are:
- add Native support (PR is on review, and it's a huge change)
- add iOS support (@CherryPerry on it)
- cover existing functionality with tests (feel free to contribute)

Lamartio:
Got it! The PR I sent now are the ones I often use and need in Android development, so was not planning to send any more operators ;)

Next I will work on a TestObserver, TestScheduler and will write some tests.

arkivanov:
Please check out these PRs:
https://github.com/badoo/Reaktive/pull/34 - here I have converted existing code to immutable (no locks and no mutable data), it makes possible to implement multi-threaded Native. Also added Single.blockingGet().
https://github.com/badoo/Reaktive/pull/36 - some tests and TestObservableObserver
I would recommend to read first PR (or wait till it will be merged), as it's a completely new code base. Cheers.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/WithLatestFrom.kt
```diff
@@ -0,0 +1,82 @@
...
+import com.badoo.reaktive.base.ErrorCallback
+import com.badoo.reaktive.completable.CompletableCallbacks
+import com.badoo.reaktive.disposable.CompositeDisposable
+import com.badoo.reaktive.disposable.Disposable
+import com.badoo.reaktive.utils.Uninitialized
+import com.badoo.reaktive.utils.lock.newLock
+import com.badoo.reaktive.utils.lock.synchronized
+import kotlin.jvm.JvmName
+
+fun <T, O, R> Observable<T>.withLatestFrom(
```
arkivanov: I believe it would be good to implement firstly `fun <T, U, R> Observable<T>.withLatestFrom(others: Collection<Observable<U>>, mapper: (T, List<U>) -> R): Observable<R>` and then just add a number of specific operators (with one, two, etc. other sources).

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/WithLatestFrom.kt
```diff
@@ -0,0 +1,82 @@
...
+                            }
+
+                        }
+                    )
+                }
+            }
+        )
+    }
+
+private class Proxy<T, O, R>(
```
arkivanov: Better to name private classes like "WithLatestFromProxy", otherwise you can't have more than one class called `Proxy` in the same package.

reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/WithLatestFromTest.kt
```diff
@@ -0,0 +1,38 @@
...
+
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class WithLatestFromTest {
+
+    private val combine: (Int, String) -> String = { value, other -> value.toString() + other }
+
+    @Test
+    fun combine() {
```
arkivanov: Please provide more meaningful names for tests

arkivanov:
Would you mind to fix your PRs so we can merge it and I'll be able to update "native" PR accordingly?

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/WithLatestFrom.kt
Lamartio: Although they share the same operator name, that is technically a `WithLatestFromMany` and requires a different implementation (see the [RxJava source](https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/internal/operators/observable/ObservableWithLatestFromMany.java)). If I were to create it, I would like to do that in another PR.

arkivanov: I'm not sure why is it like this in RxJava2 for `withLatestFrom` with one other source, but for `withLatestFrom` with two or more "others" they are using generic version of withLatestFrom (with array of sources).

We can do better, this should work :
```
fun <T, U, R> Observable<T>.withLatestFrom(others: Collection<Observable<U>>, mapper: (T, List<U>) -> R): Observable<R> =
    ...

fun <T, U, R> Observable<T>.withLatestFrom(other: Observable<U>, combine: (value: T, other: U) -> R): Observable<R> =
    withLatestFrom(listOf(other)) { item, others ->
        combine(item, others[0])
    }
```

Lamartio: Indefinitely it will work, but when comparing the two ("other" vs. "others") it appears that the "others" variant requires more (expensive) logic. Also it feels inefficient to wrap "other" in a list and later unwrap its result to call the `combine` callback. 

Although opinionated: The "others" variant is not so valuable, because in practice it rarely happens that there are "others" and when it is, their type differ from each other.

Therefore I propose we develop the "others" variant in a different PR.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/WithLatestFrom.kt
```diff
@@ -0,0 +1,82 @@
...
+    fun onNext(other: U) {
+        lock.synchronized {
+            this.other = other
+        }
+
+        emit()
+    }
+
+    @Suppress("UNCHECKED_CAST")
+    private fun emit() {
```
arkivanov: Race condition: both `onNext` methods might be called from different threads at the same time. As a result `combine` and `onNext` callbacks might be called concurrently, they should be called synchronously.

arkivanov: Maybe it's better to postpone this PR unless https://github.com/badoo/Reaktive/pull/34 will be merged. Full refactoring will be required.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/WithLatestFrom.kt
```diff
@@ -0,0 +1,82 @@
...
+        emit()
+    }
+
+    @JvmName("onNextOther")
+    fun onNext(other: U) {
+        lock.synchronized {
+            this.other = other
+        }
+
+        emit()
```
arkivanov: "Other" sources must not trigger emissions, please check [withLatestFrom](https://rxmarbles.com/#withLatestFrom).

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/WithLatestFrom.kt
Lamartio: No problem

arkivanov:
We now have withLatestFrom operator, added it as part of https://github.com/badoo/Reaktive/pull/34 to save time and release it with next alpha, closing this PR

PR #40, opened 2019-05-06
Merge our gitignore with the RxJava his gitignore

Lamartio:
When developing I found that `.DS_Store` and `out/` were able to commit. Now I merged the gitignore of RxJava, but an alternative could be to use [gitignore.io](https://gitignore.io/).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@
 .idea
 /build
 /classes
+/out
+.DS_Store*


CherryPerry:
Hi, don't you think that such `.gitignore` is an overkill? We don't have any packages produced by out builds, compiled source is in already ignored `build` and `out` folder. The whole `.idea` folder with project files is ignored too. We don't use PMD, it is a Kotlin project. Neither NetBeans nor Eclipse. Maybe just add `.DS_Store` and `out/`?

Lamartio:
Could be ;) Usually I pull in a template, but a more incremental way is also fine

.gitignore
```diff
@@ -4,3 +4,5 @@
 .idea
 /build
 /classes
+out/
```
arkivanov: I'm not sure but shouldn't it be "/out"?

Lamartio: I picked this one from a template and never had troubles with it before. As for this project, I think both choices are fine.

CherryPerry: @arkivanov is right. Tried `out/` instead of `/out' and `com.badoo.reaktive.out.Test` is ignored.

PR #42, opened 2019-05-09
Added build folders in first level subfolders to gitignore

arkivanov:


CherryPerry:
Do you want to remove `.girignore` files from subprojects? They contain only `/build` rule.

arkivanov:
> Do you want to remove `.girignore` files from subprojects? They contain only `/build` rule.

Removed

PR #47, opened 2019-05-09
Tests for Serializer

arkivanov:


reaktive/src/commonTest/kotlin/com/badoo/reaktive/utils/serializer/SerializerTest.kt
```diff
@@ -0,0 +1,171 @@
...
+
+        serializer.accept(3)
+        serializer.accept(2)
+        serializer.accept(1)
+
+        assertValues(3, 2, 1)
+    }
+
+    @Test
+    fun value_enqueued_WHEN_sent_from_callback() {
```
CherryPerry: What do we check with this test? That `onValue` callback was invoked and `success` was set true? We verify callback invocation in previous tests through `assertValues`.
You probably can't test directly, that `2` was added to queue. You can only do it by verifying emitted values like it's done in the next test.

arkivanov: I want to verify that when you are submitting a new value while processing another, it doesn't call onValue recursively.

arkivanov: And there is a separate test to verify that such a value is processed eventually.

arkivanov: I will rename the test

arkivanov: @CherryPerry renamed

PR #50, opened 2019-05-11
Tests for CombineLatest plus bug fix

arkivanov:


reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/CombineLatestTest.kt
```diff
@@ -0,0 +1,156 @@
...
+    fun erred_WHEN_not_all_sources_produced_values_and_one_source_erred_after_it_produced_a_value() {
+        emitter1.onNext("1")
+        emitter2.onNext("2")
+        emitter2.onError(Throwable())
+
+        assertFalse(observer.isCompleted)
+    }
+
+    @Test
+    fun erred_WHEN_all_sources_produced_values_and_one_source_erred() {
```
mk-jones: erred -> errored

PR #56, opened 2019-05-14
Separate implementations of Serializer and BufferedExecutor

arkivanov:
Currently both Serializer and BufferedExecutor are implemented in Common sources set and they are fully immutable. This is a huge overhead for JVM that causes a lot of GC. And same for JS, there is no need for immutability. It's better to have separate implementations of this classes for JVM, JS and Native.


reaktive/src/jsMain/kotlin/com/badoo/reaktive/scheduler/BufferedExecutor.kt
```diff
@@ -0,0 +1,30 @@
+package com.badoo.reaktive.scheduler
+
+internal actual class BufferedExecutor<in T> actual constructor(
```
mk-jones: As this is so similar to what we have in JvmCommon it may make sense to keep it in common.

Instead we could rename it to something like MutableBufferedExecutor and here plug in the mutable version. Then have AtomicBufferedExecutor for native.

reaktive/src/jsMain/kotlin/com/badoo/reaktive/utils/queue/ArrayQueue.kt
```diff
@@ -0,0 +1,74 @@
+package com.badoo.reaktive.utils.queue
+
+internal class ArrayQueue<T> : Queue<T> {
```
mk-jones: As perviously mentioned, would be nice to refactor in the future.
I find the head / tail confusing.

reaktive/src/jsMain/kotlin/com/badoo/reaktive/utils/queue/PriorityQueue.kt
```diff
@@ -0,0 +1,21 @@
...
+    private val comparator: Comparator<in T>
+) : Queue<T> {
+
+    private val list: MutableList<T> = ArrayList()
+    override val peek: T? get() = list.getOrNull(0)
+    override val size: Int get() = list.size
+
+    override fun offer(item: T) {
+        list.add(item)
+        list.sortWith(comparator) // TODO: Optimise later
```
mk-jones: :)

reaktive/src/jsMain/kotlin/com/badoo/reaktive/utils/queue/ArrayQueue.kt
arkivanov: It's a common implementation of array queue, almost copy-paste from Java ArrayDeque.

reaktive/src/jsMain/kotlin/com/badoo/reaktive/scheduler/BufferedExecutor.kt
arkivanov: Fixed, please check. The main idea is that a `sourceSet` can depend on multiple `sourceSets`. Currently this feature is not used but might be used in future. E.g. `jsMain` might depend on both `jvmJsCommon` and `jsNativeCommon` (`jsMain` -> `jvmJsCommon` -> `common` and at the same time `jsMain` -> `jsNativeCommon` -> `common`).

PR #59, opened 2019-05-15
Tests for throttle

arkivanov:


reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/ThrottleTest.kt
```diff
@@ -0,0 +1,110 @@
...
+        setTime(100L)
+        observer.reset()
+
+        emit(2)
+
+        assertValues(2)
+    }
+
+    @Test
+    fun emits_correct_values_WHEN_complex_series() {
```
mk-jones: Not sure what this is testing.

arkivanov: It sends some tricky series of values through upstream and verifies that output is correct

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/Throttle.kt
```diff
@@ -5,21 +5,25 @@ import com.badoo.reaktive.disposable.DisposableWrapper
 import com.badoo.reaktive.utils.atomicreference.AtomicReference
 import com.badoo.reaktive.utils.uptimeMillis
 
-fun <T> Observable<T>.throttle(windowMillis: Long): Observable<T> =
+private val getTimeMillis = ::uptimeMillis
+
+fun <T> Observable<T>.throttle(windowMillis: Long): Observable<T> = throttle(windowMillis, getTimeMillis)
```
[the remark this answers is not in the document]
arkivanov: Name is unchanged, it is named according to ReactiveX, `throttleFirst` is just an alias in RxJava.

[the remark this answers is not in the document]
arkivanov: Btw, `throttleLast` is the RxJava's alias for `sample` which we already have.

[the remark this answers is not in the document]
mk-jones: Cool. It's fine then.

reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/ThrottleTest.kt
mk-jones: I think it would be good to be explicit about how it is complex. 
In the future we may want tests to cover specific complex scenarios, it may mix us up what is and isn't already covered.

arkivanov: @mk-jones I can't find any words to describe this particular case :-)

PR #61, opened 2019-05-15
Add defaultIfEmpty and switchIfEmpty operator on Observable

pietrocaselani:
This PR adds the operator `defaultIfEmpty`, which is also available on other [implementations of Rx](http://reactivex.io/documentation/operators/defaultifempty.html)

I will probably create another PR adding the `switchIfEmpty`, is that ok?
Both have the same behavior, but `switchIfEmpty` receives another Observable as parameter instead of a value.

### DIFF
--- a/reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/SwitchIfEmpty.kt
+++ b/reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/SwitchIfEmpty.kt
@@ -1,0 +1,34 @@
+package com.badoo.reaktive.observable
+
+import com.badoo.reaktive.disposable.Disposable
+import com.badoo.reaktive.disposable.DisposableWrapper
+import com.badoo.reaktive.utils.atomicreference.AtomicReference
+
+fun <T> Observable<T>.switchIfEmpty(otherObservable: Observable<T>): Observable<T> =
+    observableUnsafe { observer ->
+        val disposableWrapper = DisposableWrapper()
+        observer.onSubscribe(disposableWrapper)
+
+        subscribeSafe(
+            object : ObservableObserver<T>, ObservableCallbacks<T> by observer {
+                private val isEmpty = AtomicReference(true)
+
+                override fun onSubscribe(disposable: Disposable) {
+                    disposableWrapper.set(disposable)
+                }
+
+                override fun onNext(value: T) {
+                    isEmpty.value = false
+                    observer.onNext(value)
+                }
+
+                override fun onComplete() {
+                    if (isEmpty.value) {
+                        otherObservable.subscribe(this)
+                    } else {
+                        observer.onComplete()
+                    }
+                }
+            }
+        )
+    }
--- a/reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/SwitchIfEmpty.kt
+++ b/reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/SwitchIfEmpty.kt
@@ -9,12 +9,12 @@ fun <T> Observable<T>.switchIfEmpty(otherObservable: Observable<T>): Observable<
     switchIfEmpty { otherObservable }
 
 fun <T> Observable<T>.switchIfEmpty(otherObservable: () -> Observable<T>): Observable<T> =
-    observableUnsafe { observer ->
+    observable { emitter ->
         val disposableWrapper = DisposableWrapper()
-        observer.onSubscribe(disposableWrapper)
+        emitter.setDisposable(disposableWrapper)
 
         subscribeSafe(
-            object : ObservableObserver<T>, ObservableCallbacks<T> by observer {
+            object : ObservableObserver<T>, ErrorCallback by emitter {
                 private val isEmpty = AtomicReference(true)
 
                 override fun onSubscribe(disposable: Disposable) {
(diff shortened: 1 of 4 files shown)


arkivanov:
Hey, thanks for contributions! `defaultIfEmpty` should use `switchIfEmpty` internally, so let's start with 'switchIfEmpty`. Please refer to Rxjava2 [Observable.defaultIfEmpty](https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/Observable.java#L7425) operator.

pietrocaselani:
It's ok to add into this PR or should I created another one?

arkivanov:
> It's ok to add into this PR or should I created another one?

Sure, update directly in this PR

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/SwitchIfEmpty.kt
```diff
@@ -0,0 +1,34 @@
...
+                }
+
+                override fun onNext(value: T) {
+                    isEmpty.value = false
+                    observer.onNext(value)
+                }
+
+                override fun onComplete() {
+                    if (isEmpty.value) {
+                        otherObservable.subscribe(this)
```
arkivanov: 1. Please use 'subscribeSafe'
2. You are subscribing with the same observer, if second observable will be empty as well, then you will subscribe to it again. You should create a new observer and delegate only 'Observer' and 'ErrorCallback' to 'this'. Please cover this case with tests.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/SwitchIfEmpty.kt
```diff
@@ -0,0 +1,34 @@
+package com.badoo.reaktive.observable
+
+import com.badoo.reaktive.disposable.Disposable
+import com.badoo.reaktive.disposable.DisposableWrapper
+import com.badoo.reaktive.utils.atomicreference.AtomicReference
+
+fun <T> Observable<T>.switchIfEmpty(otherObservable: Observable<T>): Observable<T> =
```
arkivanov: Let's also add 'switchIfEmpty' with lambda supplying an Observable.

pietrocaselani:
Hi @arkivanov
I not sure if I understand correctly, but I added a test as you suggested and looks good.
Thanks for the feedback!

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/SwitchIfEmpty.kt
```diff
@@ -0,0 +1,52 @@
...
+    switchIfEmpty { otherObservable }
+
+fun <T> Observable<T>.switchIfEmpty(otherObservable: () -> Observable<T>): Observable<T> =
+    observableUnsafe { observer ->
+        val disposableWrapper = DisposableWrapper()
+        observer.onSubscribe(disposableWrapper)
+
+        subscribeSafe(
+            object : ObservableObserver<T>, ObservableCallbacks<T> by observer {
```
arkivanov: You are overriding `onNext` and `onComplete` so you can replace `ObservableCallbacks<T>` with just `ErrorCallback`.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/SwitchIfEmpty.kt
```diff
@@ -0,0 +1,52 @@
...
+
+                override fun onNext(value: T) {
+                    isEmpty.value = false
+                    observer.onNext(value)
+                }
+
+                override fun onComplete() {
+                    if (isEmpty.value) {
+                        otherObservable().subscribeSafe(
+                            object : ObservableObserver<T>, ErrorCallback by this {
```
arkivanov: Please use delegation as follows:
```
object : ObservableObserver<T>, ObservableCallbacks<T> by observer {
    // override only onSubscribe here
}
```

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/SwitchIfEmpty.kt
```diff
@@ -0,0 +1,52 @@
...
+                }
+
+                override fun onNext(value: T) {
+                    isEmpty.value = false
+                    observer.onNext(value)
+                }
+
+                override fun onComplete() {
+                    if (isEmpty.value) {
+                        otherObservable().subscribeSafe(
```
arkivanov: You should try-catch client side callbacks (`otherObservable`). Something like this:
```
val nextObservable = 
    try {
        otherObservable()
    } catch (e: Throwable) {
        observer.onError(e)
        return
    }

nextObservable.subscribeSafe(
    // observer here
)
```

Also please add a test to cover the case when `otherObservable()` throws an exception.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/SwitchIfEmpty.kt
```diff
@@ -0,0 +1,52 @@
...
+import com.badoo.reaktive.base.ErrorCallback
+import com.badoo.reaktive.disposable.Disposable
+import com.badoo.reaktive.disposable.DisposableWrapper
+import com.badoo.reaktive.utils.atomicreference.AtomicReference
+
+fun <T> Observable<T>.switchIfEmpty(otherObservable: Observable<T>): Observable<T> =
+    switchIfEmpty { otherObservable }
+
+fun <T> Observable<T>.switchIfEmpty(otherObservable: () -> Observable<T>): Observable<T> =
+    observableUnsafe { observer ->
```
arkivanov: In this case it's better to use `observable { emitter ->` as if `otherObservable` will throw an exception and upstream will emit an item, then this item will be emitted to downstream (emission after error). Emitter is safe and will protect from such cases.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/SwitchIfEmpty.kt
```diff
@@ -0,0 +1,52 @@
...
+        val disposableWrapper = DisposableWrapper()
+        observer.onSubscribe(disposableWrapper)
```
arkivanov: When you convert it to `emitter` then just call `emitter.setDisposable` instead of `observer.onSusbcribe`

pietrocaselani:
@arkivanov could you please review it again?

Issue #62, opened 2019-05-16
Lock and Condition are broken in Native


arkivanov:
Latest release is still fine. Was broken in https://github.com/badoo/Reaktive/pull/51.
Affects Native schedulers, jobs might not be executed or any other unpredictable behaviour might occur.

PR #63, opened 2019-05-17
Fixed broken Native Lock and Condition

arkivanov:
Well, after further investigations it appears to be impossible to avoid manual memory freeing when using `cinterop` from Kotlin. The [implementation](https://github.com/ktorio/ktor/blob/master/ktor-utils/posix/src/io/ktor/util/LockNative.kt) from `ktor` does not work either. I have raised an [issue](https://github.com/ktorio/ktor/issues/1135) for them. Reverted the commit fafa90e1ec82b896148eb30748235d54d89123ec and updated the code a bit.
Fixes https://github.com/badoo/Reaktive/issues/62

### DIFF
--- a/reaktive/src/commonMain/kotlin/com/badoo/reaktive/single/BlockingGet.kt
+++ b/reaktive/src/commonMain/kotlin/com/badoo/reaktive/single/BlockingGet.kt
@@ -1,56 +1,60 @@
 package com.badoo.reaktive.single
 
 import com.badoo.reaktive.disposable.Disposable
-import com.badoo.reaktive.utils.Lock
 import com.badoo.reaktive.utils.atomicreference.AtomicReference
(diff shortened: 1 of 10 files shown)


reaktive/src/commonMain/kotlin/com/badoo/reaktive/single/BlockingGet.kt
```diff
@@ -1,56 +1,60 @@
...
-            }
-
-            override fun onSuccess(value: T) {
-                lock.synchronized {
-                    result.value = BlockingGetResult.Success(value)
-                    condition.signal()
+import com.badoo.reaktive.utils.useCondition
+import com.badoo.reaktive.utils.useLock
+
+fun <T> Single<T>.blockingGet(): T =
```
mk-jones: We should mark this method as not working on JS.

PR #74, opened 2019-05-23
Moved test utils to reaktive-test module

arkivanov:
- Moved test utils to a separate multiplatform module "reaktive-test"
- Disabled JS tests for a while as they are failing, @CherryPerry please fix them after :-)


reaktive-test/src/commonTest/kotlin/com/badoo/reaktive/test/DummyTest.kt
```diff
@@ -0,0 +1,10 @@
+package com.badoo.reaktive.test
+
+import kotlin.test.Test
+
+class DummyTest {
```
CherryPerry: DummyTest?

reaktive-test/build.gradle
```diff
@@ -0,0 +1,53 @@
...
+
+if (target.isCommon() || target.isMeta()) {
+    setupCommonTargets(project)
+
+    kotlin {
+        sourceSets {
+            jvmMain.dependsOn commonMain
+            jvmTest.dependsOn commonTest
+            
+            androidMain.dependsOn commonMain
```
CherryPerry: Do we need android for this project? Just publish jvm version.

reaktive-test/src/commonTest/kotlin/com/badoo/reaktive/test/DummyTest.kt
arkivanov: There was some problem with JS tests as there were no tests in module. Removed DummyTest as JS tests are switched off for now. You will probable need to add it back.

reaktive-test/build.gradle
arkivanov: I'm not sure, I don't see any downsides of this. I wanted to keep consistency. Also later we might add some android specific code and it might be a braking change perhaps. Not sure, what do you guys think?

PR #75, opened 2019-05-23
Introduce generic Source interface

arkivanov:
Common generic interface `Source` for `Observable`, `Single`, `Maybe` and `Completable`


arkivanov:
Guys, you can review just the first commit, it's very small one. Second commit is just a refactoring - imports are changed.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/utils/Utils.kt
```diff
@@ -7,12 +7,12 @@ internal fun handleSourceError(error: Throwable, onError: ((Throwable) -> Unit)?
         } else {
             try {
                 onError(error)
-            } catch (ignored: Throwable) {
-                reaktiveUncaughtErrorHandler(error)
```
mk-jones: I think it may be worth stating here that it is an exception on trying to deliver an error result. In an effort to make debugging a bit clearer.

arkivanov: Good point, fixed

reaktive/src/jsMain/kotlin/com/badoo/reaktive/utils/ThrowableExt.kt
```diff
@@ -0,0 +1,5 @@
+package com.badoo.reaktive.utils
+
+actual fun Throwable.printStack() {
+    // no-op
```
mk-jones: I had never noticed that there was no printStackTrace available for JS. Could we print the Throwable message field, then the cause field (if not null)?

arkivanov: Found a way of getting stack trace for exceptions. Check :-)

lukaville: Have you tried to just print the throwable? Probably it will display interactive error object in the log: 
image

arkivanov: If you print via `println` then it prints just class name and message. If you print via `console.log` then it provides that interactive thing. Should we use `console`? As far as I see `println` has some checks for `NodeJs` and prints either using `NodeJs` or via `console`.
image

arkivanov: @lukaville

lukaville: @arkivanov looks like in node `console.log`/`console.error` work as expected. So I think it is ok to use `console` object:
```
➜  ~ node
> console.log(Error('test'))
Error: test
    at repl:1:13
    at Script.runInThisContext (vm.js:96:20)
    at REPLServer.defaultEval (repl.js:329:29)
    at bound (domain.js:396:14)
    at REPLServer.runBound [as eval] (domain.js:409:12)
    at REPLServer.onLine (repl.js:642:10)
    at REPLServer.emit (events.js:187:15)
    at REPLServer.EventEmitter.emit (domain.js:442:20)
    at REPLServer.Interface._onLine (readline.js:290:10)
    at REPLServer.Interface._line (readline.js:638:8)
```

arkivanov: Alright will fix

arkivanov: @lukaville Fixed

reaktive/src/jsMain/kotlin/com/badoo/reaktive/utils/ThrowableExt.kt
```diff
@@ -1,18 +1,5 @@
...
-
-private val Throwable.stackTrace: String
-    get() =
-        try {
-            asDynamic().stack.toString()
-        } catch (ignored: Throwable) {
-            toString()
-        }
\ No newline at end of file
+    console.log(this)
```
lukaville: maybe let's use console.error(this)?

arkivanov: I wanted to make it consistent with other platforms. Regular printStackTrace prints via regular output in regular color. Our uncaught error handlers are printing via error.

arkivanov: @lukaville Fixed error printing, via error output

PR #78, opened 2019-05-25
Annotations in separate module

arkivanov:
- Introduced `reaktive-annotations` module
- Moved `@UseReturnValue` there
- Added `@EventsOnXxxScheduler` annotations


reaktive-annotations/src/commonMain/kotlin/com/badoo/reaktive/annotations/Annotations.kt
```diff
@@ -0,0 +1,43 @@
...
+@Retention(AnnotationRetention.SOURCE)
+@Target(AnnotationTarget.FUNCTION)
+annotation class UseReturnValue
+
+/**
+ * Indicates that annotated source does not operate on any particular scheduler
+ */
+@Retention(AnnotationRetention.SOURCE)
+@Target(AnnotationTarget.FUNCTION, AnnotationTarget.FIELD, AnnotationTarget.PROPERTY)
+annotation class EventsOnUnspecifiedScheduler
```
CherryPerry: Why? RxJava implementation allow us to use any scheduler for scheduled operations (interval, debounce and so on). You want to lock some particular operators to particular scheduler?

arkivanov: This kinda we should discuss, perhaps. The idea is that you can annotate a method or a field in e.g. a data source so you clearly see on which thread it emits. So you can decide should you switch scheduler to e.g. main or not. Also maybe we can even introduce some lint checks later.

Use case:
```kotlin
interface UserPersistentDataSource {
    @EventsOnIoScheduler
    fun load(userId: String): Single<User?>
}

interface UserNetworkDataSource {
    @EventsOnMainScheduler
    fun load(userId: String): Single<User>
}

@EventsOnMainScheduler
fun loadUser(userId: String): Single<User> =
    userPersistentDataSource
        .load(userId)
        .flatMap {
            // IO thread here
            it
                ?.toSingle()
                ?.observeOn(mainScheduler) // Need to switch to main thread
                ?: userNetworkDataSource.load(userId) // No need to switch, will be on main thread
        }
```

arkivanov: @lukaville @mk-jones Guys please comment your thoughts, this is kinda we already have in Badoo, I find it useful

CherryPerry: The idea is fine, but we have 2-way control flow for threads. What will happen when `UserPersistentDataSource.load(id).subscribeOn(mainScheduler)` appears in chain? Nobody knows, but consumers will expect results on io thread. Probably it will make sense with ability to freeze subscription on particular thread like `subscribeOnlyOn(ioScheduler)` or `verifyScheduler(ioScheduler)`. Don't know if it is possible to implement.

arkivanov: Nope, you are as a developer of data source (or any other component) can annotate its methods with these annotations depending on actual implementations. If you always produce items on a particular thread (you have `observeOn` or `subscribeOn` in your implementation) then you can annotate. If you have neither `observeOn` nor `subscribeOn` then probably `EventsOnUnspecifiedScheduler` would be useful. And so on.
In case of `UserPersistentDataSource.load(id).subscribeOn(mainScheduler)`, you should use 'subscribeOn(ioScheduler)` in your implementation so you always will work on IO thread and `subcribeOn(mainScheduler)` does not matter.

arkivanov: So e.g. the following function can be annotated (it's scheduler is guaranteed):
```kotlin
@EventsOnIoScheduler
fun loadUser(): Single<User> =
        single<User> {
            // Load user from database
        }
            .subscribeOn(ioScheduler)

```

mk-jones: I like this. It's convenient to know for the api client. Though I would have called it EmitsOnXXX

arkivanov: > I like this. It's convenient to know for the api client. Though I would have called it EmitsOnXXX

I also would love "EmitsOnXxx" but it's not really applicable to `Completable`, as it does not emit. But it is still useful in cases: `Completable.andThen(...)`, `Maybe.switchIfEmpty(...)`, etc.

arkivanov:
> I think EmitsOnXXX is a better name as it is more specific about results being sent over that scheduler. Client devs may confuse EventsXXX with where operation is performed. Overall I think the idea is good, it is a helpful thing to know when working with an Observer pattern api.

Please read my comment above and let me know your thoughts :-)

arkivanov:
@lukaville could you please share your thoughts as well? 😀

reaktive-annotations/src/commonMain/kotlin/com/badoo/reaktive/annotations/Annotations.kt
lukaville: ➕ 
let's add these annotations

arkivanov:
@CherryPerry Should I add/register the new module somewhere?

CherryPerry:
@arkivanov You can add this module to check list for CI [here](https://github.com/badoo/Reaktive/blob/master/.travis.yml#L11), but it is not mandatory because this module contains only annotations.

Also we need check publications later before next release because of new modules.

PR #79, opened 2019-05-25
Tests for Single.flatMap

arkivanov:
- Introduced base classes `TestSource` and `TestObserver`
- Refactored `TestObservable` and `TestObservableObserver`
- Introduced `TestSingle` and `TestSingleObserver`
- Added tests for `Single.flatMap`


reaktive-test/src/commonMain/kotlin/com/badoo/reaktive/test/base/TestObserverExt.kt
```diff
@@ -0,0 +1,9 @@
+package com.badoo.reaktive.test.base
+
+import com.badoo.reaktive.disposable.Disposable
+
+fun TestObserver<*>.dispose() {
```
CherryPerry: Just make `TestObserver` disposable?

PR #86, opened 2019-05-28
Tests for SingleByEmitter

arkivanov:


reaktive/src/commonTest/kotlin/com/badoo/reaktive/single/SingleByEmitterTest.kt
```diff
@@ -0,0 +1,112 @@
...
+
+    @Test
+    fun completed_with_error_WHEN_exception_during_subscribe() {
+        val error = RuntimeException()
+
+        single<Int> { throw error }.subscribe(observer)
+
+        assertTrue(observer.isError(error))
+    }
+}
```
CherryPerry: Add tests for `setDisposable` please.

arkivanov: Fixed

CherryPerry: Not sure about it.

If emitter's disposable is disposed on dispose call, then why do we need calling in manually at [SingleByEmitter.kt#L18](https://github.com/badoo/Reaktive/blob/master/reaktive/src/commonMain/kotlin/com/badoo/reaktive/single/SingleByEmitter.kt#L18)?

If it is already called, then remove manual call from L18. If it is not, then cover L18 case too.

arkivanov: There are multiple cases when disposable should be disposed:
- when Emitter is signalled with `onSuccess`, `onComplete` or `onError` (depending on its type)
- when downstream disposes a disposable provided by the upstream (the Single under test)

CherryPerry: Then add `assigned_disposable_is_disposed_WHEN_onSuccess` test.

arkivanov: Added missing tests for `SingleByEmitter` and `ObservableByEmitter`. Two tests are now failing because of the bug in `DisposableWrapper` (https://github.com/badoo/Reaktive/issues/91).
@CherryPerry Meanwhile you can review the changes here.

PR #90, opened 2019-05-30
Single.filter and Single.notNull

CherryPerry:
Both of them transforms `Single` to `Maybe`.
Also we should decide, how to implement `UpstreamDownstreamGenericTests` for cases, when we transform one type of stream to another.

### DIFF
--- a/reaktive/src/commonMain/kotlin/com/badoo/reaktive/single/Filter.kt
+++ b/reaktive/src/commonMain/kotlin/com/badoo/reaktive/single/Filter.kt
@@ -4,13 +4,22 @@ import com.badoo.reaktive.base.ErrorCallback
 import com.badoo.reaktive.base.Observer
 import com.badoo.reaktive.base.subscribeSafe
 import com.badoo.reaktive.base.tryCatch
+import com.badoo.reaktive.disposable.Disposable
+import com.badoo.reaktive.disposable.DisposableWrapper
 import com.badoo.reaktive.maybe.Maybe
 import com.badoo.reaktive.maybe.maybeUnsafe
 
 fun <T> Single<T>.filter(predicate: (T) -> Boolean): Maybe<T> =
     maybeUnsafe { observer ->
+        val disposableWrapper = DisposableWrapper()
+        observer.onSubscribe(disposableWrapper)
+
         subscribeSafe(
             object : SingleObserver<T>, Observer by observer, ErrorCallback by observer {
+                override fun onSubscribe(disposable: Disposable) {
+                    disposableWrapper.set(disposable)
+                }
+
                 override fun onSuccess(value: T) {
                     observer.tryCatch({ predicate(value) }) {
                         if (it) {
--- a/reaktive/src/commonMain/kotlin/com/badoo/reaktive/single/Filter.kt
+++ b/reaktive/src/commonMain/kotlin/com/badoo/reaktive/single/Filter.kt
@@ -1,0 +1,25 @@
+package com.badoo.reaktive.single
+
+import com.badoo.reaktive.base.ErrorCallback
+import com.badoo.reaktive.base.Observer
+import com.badoo.reaktive.base.subscribeSafe
+import com.badoo.reaktive.base.tryCatch
+import com.badoo.reaktive.maybe.Maybe
+import com.badoo.reaktive.maybe.maybeUnsafe
+
+fun <T> Single<T>.filter(predicate: (T) -> Boolean): Maybe<T> =
+    maybeUnsafe { observer ->
+        subscribeSafe(
+            object : SingleObserver<T>, Observer by observer, ErrorCallback by observer {
+                override fun onSuccess(value: T) {
+                    observer.tryCatch({ predicate(value) }) {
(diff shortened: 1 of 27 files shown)


arkivanov:
@CherryPerry You missed yesterday meeting :-) All agreed that we can remove Maybe in favour of Single<T?>. We just wanted your opinion. Looks like there is no point to maintain whole Maybe. We could add kinda `Single.switchIfNull()` if we really want.

CherryPerry:
@arkivanov `Maybe` is cool. If you replace it with `Single` then all your chain will be corrupted with nullability checks (e.g. `single { null }.map { it?.length }.map { it?.plus(1) }`). You can obviously switch back from `Single<T?>` to `Observable<T>`, but it breaks single result contract. `Single.switchIfNull()` just pollutes your codebase with `SomeClass.EMPTY` or `SomeClass.NULL` objects as it is now in RxJava2.

arkivanov:
> Single.switchIfNull() just pollutes your codebase with SomeClass.EMPTY or SomeClass.NULL objects as it is now in RxJava2

I didn't really get the point, could you elaborate?

CherryPerry:
@arkivanov How does `Single.switchIfNull()` work?

reaktive/src/commonMain/kotlin/com/badoo/reaktive/single/Filter.kt
```diff
@@ -0,0 +1,25 @@
...
+import com.badoo.reaktive.base.ErrorCallback
+import com.badoo.reaktive.base.Observer
+import com.badoo.reaktive.base.subscribeSafe
+import com.badoo.reaktive.base.tryCatch
+import com.badoo.reaktive.maybe.Maybe
+import com.badoo.reaktive.maybe.maybeUnsafe
+
+fun <T> Single<T>.filter(predicate: (T) -> Boolean): Maybe<T> =
+    maybeUnsafe { observer ->
+        subscribeSafe(
```
arkivanov: `subscribeSafe` will call `observer.onError` in case of exception during subscription. No callbacks should be called before `obsrever.onSubscribe`. Use `DisposableWrapper` trick. Same mistake I made in `Maybe.filter`, please fix. And some tests would be good.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/single/NotNull.kt
```diff
@@ -0,0 +1,22 @@
...
+
+import com.badoo.reaktive.base.ErrorCallback
+import com.badoo.reaktive.base.Observer
+import com.badoo.reaktive.base.subscribeSafe
+import com.badoo.reaktive.maybe.Maybe
+import com.badoo.reaktive.maybe.maybeUnsafe
+
+fun <T> Single<T?>.notNull(): Maybe<T> =
+    maybeUnsafe { observer ->
+        subscribeSafe(
```
arkivanov: Same here

reaktive/src/commonTest/kotlin/com/badoo/reaktive/single/UpstreamDownstreamGenericTests.kt
```diff
@@ -41,4 +44,28 @@ interface UpstreamDownstreamGenericTests {
                 }
             }
     }
+
+    object TransformToMaybe {
```
arkivanov: It can be just a function inside existing companion object. Like `transformToMaybe(...): UpstreamDownstreamGenericTests`

arkivanov: Just realised that we also need `single-to-observable` etc. generic tests. What do you think of:
- rename everywhere `UpstreamDownstreamGenericTests` to e.g. `SingleToSingleTests`, `ObservableToObservableTests`, `SingleToMaybeTests`? Each in separate file in corresponding packages.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/single/NotNull.kt
```diff
@@ -0,0 +1,31 @@
...
+
+import com.badoo.reaktive.base.ErrorCallback
+import com.badoo.reaktive.base.Observer
+import com.badoo.reaktive.base.subscribeSafe
+import com.badoo.reaktive.disposable.Disposable
+import com.badoo.reaktive.disposable.DisposableWrapper
+import com.badoo.reaktive.maybe.Maybe
+import com.badoo.reaktive.maybe.maybeUnsafe
+
+fun <T> Single<T?>.notNull(): Maybe<T> =
```
arkivanov: Better to define type `T` as `<T: Any>`

reaktive/src/commonMain/kotlin/com/badoo/reaktive/single/Filter.kt
CherryPerry: > And some tests would be good (perhaps generic tests to cover such cases everywhere).

Not sure how to implement it, let's move it to another PR.

PR #92, opened 2019-06-01
Tests for DisposableWrapper

### LINKED ISSUE #91: DisposableWrapper does not dispose new disposable if already disposed
Scenario:
```
fun test() {
    val wrapper = DisposableWrapper()
    wrapper.dispose()
    val disposable = disposable()
    wrapper.set(disposable)
    assertTrue(disposable.isDisposed)
}
```

arkivanov:
Fixes #91

### DIFF
--- a/reaktive/src/commonMain/kotlin/com/badoo/reaktive/disposable/DisposableWrapper.kt
+++ b/reaktive/src/commonMain/kotlin/com/badoo/reaktive/disposable/DisposableWrapper.kt
@@ -35,10 +35,10 @@ class DisposableWrapper : Disposable {
             ?.dispose()
     }
 
-    private class Holder(val disposable: Disposable?)
-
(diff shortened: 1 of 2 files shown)


reaktive/src/commonMain/kotlin/com/badoo/reaktive/disposable/DisposableWrapper.kt
```diff
@@ -24,10 +25,21 @@ class DisposableWrapper : Disposable {
...
+
+                holder
+            }
             ?.dispose()
     }
 
     private class Holder(val disposable: Disposable?)
+
+    private companion object {
+        private fun Holder.dispose() {
```
CherryPerry: Why extension function? You have access to `Holder` class, create dispose function there.

PR #96, opened 2019-06-08
Tests for doOnBeforeXxx

arkivanov:
- renamed `.isCompleted` to `isComplete` (for `TestObservableObserver` and `TestCompletableObserver`)
- renamed `completable.UpstreamDownstreamGenericTests` to `CompletableToCompletableTests`
- added tests for `doOnBeforeXxx` for all sources
- found and fixed bugs in `doOnBeforeSubscribe`

_First commit is refactoring, so it's better to review by commits._

### DIFF
--- a/reaktive/src/commonTest/kotlin/com/badoo/reaktive/completable/DoOnBeforeCompleteTest.kt
+++ b/reaktive/src/commonTest/kotlin/com/badoo/reaktive/completable/DoOnBeforeCompleteTest.kt
@@ -2,16 +2,20 @@ package com.badoo.reaktive.completable
 
 import com.badoo.reaktive.test.completable.DefaultCompletableObserver
 import com.badoo.reaktive.test.completable.TestCompletable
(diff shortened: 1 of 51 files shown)


reaktive/src/commonTest/kotlin/com/badoo/reaktive/completable/DoOnBeforeCompleteTest.kt
```diff
@@ -0,0 +1,33 @@
...
+                        callOrder += "onComplete"
+                    }
+                }
+            )
+
+        upstream.onComplete()
+
+        assertEquals(listOf("action", "onComplete"), callOrder.items)
+    }
+}
```
CherryPerry: Can you add at least one test case, when "doOnBeforeXXX" should not be called? For example: do not call `onComplete` in this test and verify, that `doOnBeforeComplete` block was not called. `doOnBeforeComplete` can have broken behaviour and be executed when subscription was created. In this case test still passes.

Issue #89, opened 2019-05-29
Migrate to new API for resolving test executable for iOS


CherryPerry:
https://kotlinlang.org/docs/reference/building-mpp-with-gradle.html#building-final-native-binaries

[Current API](https://github.com/badoo/Reaktive/blob/master/build.gradle#L170) is already deprecated and will be removed in the next version.

Issue #81, opened 2019-05-26
Create "travisCheck" task


CherryPerry:
Current way of listing all tasks in `.travis.yml` is not convenient enough. It will be much better if travis will have only `./gradlew travisCheck` call.

PR #98, opened 2019-06-09
iOS plugin

CherryPerry:
- Move iOS configuration to plugin
- Remove TODO to turn on/off `check` task dependency on `iosTask`
- Make `check` depends on `iosTask` and all iOS binaries compilation tasks to verify, that lib can be built for all platforms and to simplify Travis configuration (fixes https://github.com/badoo/Reaktive/issues/81)
- Use new API to get path to executable iOS test binary (fixes https://github.com/badoo/Reaktive/issues/89)

### DIFF
--- a/buildSrc/src/main/kotlin/IosPlugin.kt
+++ b/buildSrc/src/main/kotlin/IosPlugin.kt
@@ -1,0 +1,62 @@
+import org.gradle.api.Plugin
+import org.gradle.api.Project
+import org.gradle.api.file.ConfigurableFileCollection
+import org.gradle.api.internal.AbstractTask
+import org.gradle.api.tasks.InputFiles
+import org.gradle.api.tasks.TaskAction
+import org.jetbrains.kotlin.gradle.dsl.KotlinMultiplatformExtension
+import org.jetbrains.kotlin.gradle.plugin.mpp.NativeBuildType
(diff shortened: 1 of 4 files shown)


arkivanov:
For me it's fine) @lukaville please review as well

buildSrc/src/main/kotlin/IosPlugin.kt
```diff
@@ -0,0 +1,109 @@
...
+
+        @TaskAction
+        open fun runTest() {
+            val files = testExecutables.filter { it.exists() }.files.toTypedArray()
+            if (files.isNotEmpty()) {
+                project.exec {
+                    commandLine("xcrun", "simctl", "spawn", "iPhone X", *files)
+                }
+            } else {
+                logger.error("No test executable for iOS")
```
lukaville: maybe we should throw an exception if it is impossible to run tests?

CherryPerry: We have modules without tests and I don't know a way to verify if test binary will be created or not. As I know, generate test binary task is still executable even there is no tests. It just generates nothing as output.

lukaville: :(

giphy

PR #104, opened 2019-06-12
Use hrtime in JS uptimeMillis when running in NodeJS

### LINKED ISSUE #103: uptimeMillis() fails in JS tests
`ReferenceError: window is not defined`. Should use `Date.now()`.

arkivanov:
Fixes #103

### DIFF
--- a/reaktive/src/jsMain/kotlin/com/badoo/reaktive/utils/UptimeMillis.kt
+++ b/reaktive/src/jsMain/kotlin/com/badoo/reaktive/utils/UptimeMillis.kt
@@ -1,6 +1,6 @@
 package com.badoo.reaktive.utils
 
-import kotlin.browser.window
+import kotlin.js.Date
 
 internal actual val uptimeMillis: Long
-    get() = window.performance.now().toLong()
+    get() = Date.now().toLong()
--- a/reaktive/src/jsMain/kotlin/com/badoo/reaktive/utils/UptimeMillis.kt
+++ b/reaktive/src/jsMain/kotlin/com/badoo/reaktive/utils/UptimeMillis.kt
@@ -1,6 +1,16 @@
 package com.badoo.reaktive.utils
 
-import kotlin.js.Date
+import kotlin.browser.window
 
-internal actual val uptimeMillis: Long
-    get() = Date.now().toLong()
+internal actual val uptimeMillis: Long get() = if (isWindowDefined) window.performance.now().toLong() else hrTimeMillis
+
+private external val process: dynamic
+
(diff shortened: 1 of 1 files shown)


reaktive/src/jsMain/kotlin/com/badoo/reaktive/utils/UptimeMillis.kt
```diff
@@ -1,6 +1,6 @@
 package com.badoo.reaktive.utils
 
-import kotlin.browser.window
+import kotlin.js.Date
 
 internal actual val uptimeMillis: Long
-    get() = window.performance.now().toLong()
+    get() = Date.now().toLong()
```
lukaville: What's the problem with `window.performance.now`? Looks like it's available in all modern browsers https://caniuse.com/#search=performance.now
If we use Date then uptimeMillis will not be monotonic.

CherryPerry: But not in NodeJS as I can see.
https://stackoverflow.com/questions/23003252/window-performance-now-equivalent-in-nodejs

CherryPerry: We use NodeJS to run our tests.

lukaville: Let's use https://nodejs.org/api/process.html#process_process_hrtime_time for nodejs then and window.performance.now for browsers

arkivanov:
How we can distinguish?

lukaville:
@arkivanov 
We can copy-paste from Klock or add it as a dependency https://github.com/korlibs/klock/blob/master/klock/src/jsMain/kotlin/com/soywiz/klock/internal/KlockInternal.kt

arkivanov:
Makes sense, why can't we always use hrtime?

lukaville:
> Makes sense, why can't we always use hrtime?

hrtime is not available in browsers

arkivanov:
Last question, what exactly wrong with Date?

lukaville:
> Last question, what exactly wrong with Date?

It may decrease during the execution of the program and it can lead to various unwanted side effects and broken logic especially in time-related operators

reaktive/src/jsMain/kotlin/com/badoo/reaktive/utils/UptimeMillis.kt
arkivanov: Fixed pls check

PR #106, opened 2019-06-13
Test Observable.filter

CherryPerry:


reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/FilterTest.kt
```diff
@@ -0,0 +1,61 @@
...
+import com.badoo.reaktive.test.observable.TestObservable
+import com.badoo.reaktive.test.observable.isError
+import com.badoo.reaktive.test.observable.test
+import com.badoo.reaktive.test.observable.values
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+
+class FilterTest : ObservableToObservableTests by ObservableToObservableTests<Unit>({ filter { true } }) {
```
arkivanov: Let's also add a test for series of values, e.g. `-1, null, 0, 1, 2` and assert output is `null, 1, 2`

PR #120, opened 2019-06-18
Test Completable.concat

CherryPerry:


reaktive/src/commonMain/kotlin/com/badoo/reaktive/completable/Concat.kt
```diff
@@ -1,16 +1,20 @@
...
-import com.badoo.reaktive.base.subscribeSafe
+import com.badoo.reaktive.utils.atomicreference.AtomicReference
+import com.badoo.reaktive.utils.atomicreference.getAndUpdate
 
 fun Iterable<Completable>.concat(): Completable =
     completableUnsafe { observer ->
         val disposableWrapper = DisposableWrapper()
         observer.onSubscribe(disposableWrapper)
 
-        val iterator = iterator()
```
arkivanov: Good catch! Consider just copy-pasting the existing code from `Observable.merge` to keep code more consistent.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/completable/Concat.kt
```diff
@@ -1,16 +1,20 @@
...
-        if (!iterator.hasNext()) {
+        val index = AtomicReference(0)
+        val list = toList()
```
arkivanov: Please remove `freeze()` call from `Observable.concat`, looks like it was added by mistake

PR #126, opened 2019-06-22
Kotlin version management

CherryPerry:
Moved some dependency versions to separate `versions.gradle.kts` file and set them as project extra.
In future should probably be implemented as `resolutionStrategy` for all projects.

### DIFF
--- a/buildSrc/src/main/kotlin/JsPlugin.kt
+++ b/buildSrc/src/main/kotlin/JsPlugin.kt
@@ -65,8 +65,8 @@ abstract class JsPlugin : Plugin<Project> {
             setArgs(
                 listOf(
                     "install",
-                    "kotlin@${project.findProperty("kotlin_version")}",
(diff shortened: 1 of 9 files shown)


buildSrc/src/main/kotlin/JsPlugin.kt
```diff
@@ -58,16 +58,16 @@ abstract class JsPlugin : Plugin<Project> {
         target.pluginManager.apply(NodePlugin::class.java)
         target.extensions.configure(NodeExtension::class.java) {
             download = true
-            version = Versions.node
+            version = "12.2.0"
```
arkivanov: Shouldn't we use property?

CherryPerry: I am not sure. Probably we won't check and upgrade this version, and this code will be removed when Kotlin 1.3.40 will be ready.

PR #131, opened 2019-07-03
Refactor test observers

arkivanov:
Refactored test observers, so they do checks on every event, not just accumulate. Found a few bugs and fixed them: `firstOrComplete` and `firstOrDefault`.


reaktive-test/src/commonMain/kotlin/com/badoo/reaktive/test/base/TestObserverExt.kt
```diff
@@ -0,0 +1,31 @@
+package com.badoo.reaktive.test.base
+
+import com.badoo.reaktive.test.utils.assertEquals
+import com.badoo.reaktive.test.utils.assertFalse
+import com.badoo.reaktive.test.utils.assertTrue
+
+val TestObserver.isError: Boolean get() = error != null
+
+fun TestObserver.assertError() {
```
CherryPerry: If we have access to `TestObserver` classes, why do we need extension functions? Put them inside corresponding classes to make them more visible and accessible?

reaktive-test/src/commonMain/kotlin/com/badoo/reaktive/test/utils/Assertions.kt
```diff
@@ -0,0 +1,30 @@
+/**
```
CherryPerry: Let's check it again!

reaktive-test/src/commonMain/kotlin/com/badoo/reaktive/test/base/TestObserverExt.kt
arkivanov: Discussed in person

PR #132, opened 2019-07-04
Added delay operators

arkivanov:


reaktive/src/commonMain/kotlin/com/badoo/reaktive/completable/Delay.kt
```diff
@@ -0,0 +1,39 @@
+package com.badoo.reaktive.completable
+
+import com.badoo.reaktive.base.subscribeSafe
+import com.badoo.reaktive.disposable.CompositeDisposable
+import com.badoo.reaktive.disposable.Disposable
+import com.badoo.reaktive.scheduler.Scheduler
+
+fun Completable.delay(delayMillis: Long, scheduler: Scheduler, delayError: Boolean = false): Completable =
```
CherryPerry: Do you want to add default scheduler as RxJava do?

reaktive/src/commonTest/kotlin/com/badoo/reaktive/completable/DelayTest.kt
```diff
@@ -0,0 +1,65 @@
...
+        observer.assertError()
+    }
+
+    @Test
+    fun not_delayed_error_cancels_pending_complete() {
+        upstream.onComplete()
+
+        upstream.onError(Throwable())
+
+        timer.advanceBy(1000L)
```
CherryPerry: You do not assert anything.

reaktive/src/commonTest/kotlin/com/badoo/reaktive/completable/DelayTest.kt
```diff
@@ -0,0 +1,65 @@
...
+
+        observer.assertComplete()
+    }
+
+    @Test
+    fun not_delayed_error_goes_before_pending_complete() {
+        upstream.onComplete()
+
+        upstream.onError(Throwable())
```
CherryPerry: Is it breach of contract? You can't complete and error at same time.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/completable/Delay.kt
arkivanov: Let's discuss. I find default scheduler as `computation` very error prone, as in most of the cases you want main thread. We had some crashes due to such mistakes. So I would prefer either to define default value as `mainScheduler` or to just keep it mandatory.
@CherryPerry @lukaville @mk-jones What do guys think?

reaktive/src/commonTest/kotlin/com/badoo/reaktive/completable/DelayTest.kt
arkivanov: Actually it does internally, as `TestObserver` will reject any activity after a terminal event. Perhaps I should put a comment.

reaktive/src/commonTest/kotlin/com/badoo/reaktive/completable/DelayTest.kt
arkivanov: Nope, `complete` event is delayed and if `delayErrors = false` then `error` event should cancel any pending events and go without any delay.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/completable/Delay.kt
mk-jones: I agree that it can be error prone.

I disagree with setting it mainScheduler however. I think by breaking convention with RxJava would also lead to unexpected bugs for users. I think the choice should be between providing computation as default, or to have it provided explicitly.

I would personally prefer leaving it explicit.

CherryPerry: > as in most of the cases you want main thread
In most of our cases.
Other people do a lot of staff on threads other than main. I prefer to not have any scheduler rather than have main as default.

reaktive/src/commonTest/kotlin/com/badoo/reaktive/completable/DelayTest.kt
mk-jones: It would be nice if we could explicity assert that no complete event is sent.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/completable/Delay.kt
arkivanov: Rxjava does not have main scheduler, but we do. Let's keep it explicit then.

PR #138, opened 2019-07-27
Handle exceptions in doOnBeforeXxx lambdas

### LINKED ISSUE #137: Handle exceptions in doOn*** methods
Currently exceptions thrown in doOn*** methods are not handled at all.
Sample test:
```
    @Test
    fun testErrorIsPropagated() {
        val observer = singleOf(1)
            .doOnBeforeSuccess { throw Exception("Simulated error") }
            .test()
        assertTrue(observer.isError)
    }
```
RxJava in case of exceptions in such methods proceeds with onError, so it would be great if Reaktive could support that behaviour

arkivanov:
Fixes #137


reaktive/src/commonMain/kotlin/com/badoo/reaktive/base/ErrorCallbackExt.kt
```diff
@@ -1,11 +1,15 @@
 package com.badoo.reaktive.base
 
-internal inline fun <T> ErrorCallback.tryCatch(block: () -> T, onSuccess: (T) -> Unit) {
+internal inline fun <T> ErrorCallback.tryCatch(
+    block: () -> T,
+    errorTransformer: (Throwable) -> Throwable = { it },
```
CherryPerry: The only purpose of this transformer is to create CompositeException. Can we replace it with something like
```kotlin
tryCatch(
    block: ...,
    composeExceptionWith: Throwable? = null,
    onSuccess: ...
)
```

reaktive/src/commonMain/kotlin/com/badoo/reaktive/base/exceptions/CompositeException.kt
```diff
@@ -0,0 +1,10 @@
+package com.badoo.reaktive.base.exceptions
+
+expect class CompositeException(
+    cause1: Throwable,
+    cause2: Throwable
+) : RuntimeException {
+
+    val cause1: Throwable
```
CherryPerry: Awful naming. Can we use list of throwable as RxJava do? Or maybe something like `originalException` and `errorHandlerException`?

If we use throwable list, then we can change `tryCatch()` method, so we can reduce nesting. If original exception is already `CompositeException`, then instead of creating new `CompositeException` of `CompositeException` and other exception, just add new one to list.

reaktive/build.gradle
```diff
@@ -81,4 +81,4 @@ if (target.isIos() || target.isMeta()) {
             iosSimTest.dependsOn iosCommonTest
         }
     }
-}
\ No newline at end of file
+}
```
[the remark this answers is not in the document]
arkivanov: Will revert

reaktive/src/commonMain/kotlin/com/badoo/reaktive/base/ErrorCallbackExt.kt
arkivanov: I wanted more generic function, so it could be reused later in different case. `composeExceptionWith` is very specific.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/base/exceptions/CompositeException.kt
arkivanov: The problem as always with Kotlin Native. You can't override `printStackTrace` method. So you can only print `this` exception and a `cause`. That's why I made only two fixed sub-exceptions. So `CompositeException` is the exception holding exactly two sub-exceptions: `cause1` and `cause2`. From my point of view they are on the same level, I can't say that one is major and another one is minor.
Naming like `errorHandlerException` will tie this class to the specific use case where error handling produces another error. We might want to reuse it in different ones.
What do you think?

CherryPerry: But how I as consumer will understand, which of this exceptions to use as primary source?
What about using list and create new `CompositeException` each time, passing actual exception as `cause` for constructor?

reaktive/src/commonMain/kotlin/com/badoo/reaktive/base/ErrorCallbackExt.kt
CherryPerry: Then maybe create overloaded version with `composeExceptionWith` parameter?
I don't like all this copy-pasted `{ CompositeException(it, exception) }`.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/base/exceptions/CompositeException.kt
arkivanov: Why would you need to understand where is the primary? How would you understand it with list? CompositeException with list of exceptions can not be fully printed in KN. That's why I made two fixed sub-exceptions. Please provide some examples as I don't fully understand.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/base/ErrorCallbackExt.kt
arkivanov: Will do

arkivanov: Tried, it clashes with existing version in some circumstances, e.g. if you specify only `block` and `onSuccess` arguments in the following way: `tryCatch({...}) { ... }`.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/base/exceptions/CompositeException.kt
CherryPerry: > How would you understand it with list?
The first one in list. But to find the initial error in you implementation I need to go through all nested `CompositeException`s.
> `CompositeException` with list of exceptions can not be fully printed in KN.
I am not sure, but looks like 
```kotlin
actual class CompositeException actual constructor(
    actual val cause1: Throwable,
    actual val cause2: Throwable
) : RuntimeException("Inner exception: $cause2") 
```
will print itself and `cause2`, not `cause1`.

I am fine if you don't want to change it to list now, but probably in future this API will become a subject to change.

arkivanov: Nice catch! Fixed!

arkivanov: So for `throw CompositeException(RuntimeException(), IllegalStateException())` it prints:
```
Uncaught Kotlin exception: com.badoo.reaktive.base.exceptions.CompositeException: Inner exception: kotlin.IllegalStateException
        at kfun:kotlin.RuntimeException.<init>(kotlin.String?;kotlin.Throwable?)kotlin.RuntimeException (0x22a129)
        at kfun:com.badoo.reaktive.base.exceptions.CompositeException.<init>(kotlin.Throwable;kotlin.Throwable)com.badoo.reaktive.base.exceptions.CompositeException (0x22a091)
        at kfun:com.badoo.reaktive.sample.linux.main() (0x229f7f)
        at Konan_start (0x229e81)
        at Konan_run_start (0x229e03)
        at Konan_main (0x229d77)
        at __libc_start_main (0x7fa39eb0509b)
        at  (0x223029)
        at  ((nil))
Caused by: kotlin.RuntimeException
        at kfun:kotlin.Exception.<init>()kotlin.Exception (0x225cb1)
        at kfun:kotlin.RuntimeException.<init>()kotlin.RuntimeException (0x225c81)
        at kfun:com.badoo.reaktive.sample.linux.main() (0x229f43)
        at Konan_start (0x229e81)
        at Konan_run_start (0x229e03)
        at Konan_main (0x229d77)
        at __libc_start_main (0x7fa39eb0509b)
        at  (0x223029)
        at  ((nil))

```

Unfortunately there is no way to print cause2 in full, at least I can't find a way...

PR #140, opened 2019-07-30
Use List instead of Set in CompositeDisposable

arkivanov:


arkivanov:
@CherryPerry Could you please help with the Android SDK license in Travis?

reaktive/src/commonMain/kotlin/com/badoo/reaktive/disposable/CompositeDisposable.kt
```diff
@@ -26,8 +26,11 @@ class CompositeDisposable : Disposable {
         set
             .getAndUpdate {
                 it
-                    ?.filterNotTo(hashSetOf(), Disposable::isDisposed)
-                    ?.plus(disposable)
+                    ?.toCollection(ArrayList(it.size + 1))
+                    ?.apply {
+                        removeAll(Disposable::isDisposed)
```
CherryPerry: It will work because this `ArrayList` is not frozen yet, right?

reaktive/src/commonMain/kotlin/com/badoo/reaktive/disposable/CompositeDisposable.kt
```diff
@@ -9,7 +9,7 @@ import com.badoo.reaktive.utils.atomic.getAndUpdate
  */
 class CompositeDisposable : Disposable {
 
-    private val set = AtomicReference<Set<Disposable>?>(emptySet(), true)
+    private val set = AtomicReference<List<Disposable>?>(emptyList(), true)
```
CherryPerry: Rename `set` to `list`. Why not to use `AtomicList`?

reaktive/src/commonMain/kotlin/com/badoo/reaktive/disposable/CompositeDisposable.kt
arkivanov: Yep

PR #142, opened 2019-07-31
Added Observable.concatWith and Observable.startWith(item) operators

dmitryustimov:
Basically new methods are just syntax sugar. But they provide more declarative way to describe events flow

### DIFF
--- a/reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/ConcatWith.kt
+++ b/reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/ConcatWith.kt
@@ -1,0 +1,5 @@
+package com.badoo.reaktive.observable
+
+fun <T> Observable<T>.concatWith(other: Observable<T>): Observable<T> = concat(this, other)
+
+fun <T> Observable<T>.concatWith(item: T): Observable<T> = concat(this, item.toObservable())
--- a/reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/ConcatWith.kt
+++ b/reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/ConcatWith.kt
@@ -2,4 +2,4 @@ package com.badoo.reaktive.observable
 
 fun <T> Observable<T>.concatWith(other: Observable<T>): Observable<T> = concat(this, other)
 
-fun <T> Observable<T>.concatWith(item: T): Observable<T> = concat(this, item.toObservable())
+fun <T> Observable<T>.concatWith(value: T): Observable<T> = concat(this, value.toObservable())
--- a/reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/ConcatWith.kt
+++ b/reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/ConcatWith.kt
@@ -2,4 +2,4 @@ package com.badoo.reaktive.observable
 
 fun <T> Observable<T>.concatWith(other: Observable<T>): Observable<T> = concat(this, other)
 
-fun <T> Observable<T>.concatWith(value: T): Observable<T> = concat(this, value.toObservable())
+fun <T> Observable<T>.concatWithValue(value: T): Observable<T> = concat(this, value.toObservable())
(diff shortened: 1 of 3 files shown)


sample-android-app/src/main/java/com/badoo/reaktive/sample/android/MainActivity.kt
```diff
@@ -24,11 +25,11 @@ class MainActivity : AppCompatActivity() {
 
         findViewById<Button>(R.id.button).setOnClickListener {
             observable<String> { emitter ->
-                emitter.onNext("Loading...")
                 Thread.sleep(1000L)
                 emitter.onNext(SimpleDateFormat.getDateTimeInstance().format(Date()))
                 emitter.onComplete()
             }
+                .startWith("Loading...")
```
arkivanov: Better to put it after 'observeOn', so only blocking observable will be affected by thread switching.

sample-android-app/src/main/java/com/badoo/reaktive/sample/android/MainActivity.kt
```diff
@@ -24,11 +25,11 @@ class MainActivity : AppCompatActivity() {
 
         findViewById<Button>(R.id.button).setOnClickListener {
             observable<String> { emitter ->
```
arkivanov: Also it can be replaced with 'observableFromFunction'. So just sleep and return the value.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/ConcatWith.kt
```diff
@@ -0,0 +1,5 @@
+package com.badoo.reaktive.observable
+
+fun <T> Observable<T>.concatWith(other: Observable<T>): Observable<T> = concat(this, other)
+
+fun <T> Observable<T>.concatWith(item: T): Observable<T> = concat(this, item.toObservable())
```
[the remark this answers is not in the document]
dmitryustimov: Done

sample-android-app/src/main/java/com/badoo/reaktive/sample/android/MainActivity.kt
dmitryustimov: Done

sample-android-app/src/main/java/com/badoo/reaktive/sample/android/MainActivity.kt
dmitryustimov: Done

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/ConcatWith.kt
```diff
@@ -0,0 +1,5 @@
+package com.badoo.reaktive.observable
+
+fun <T> Observable<T>.concatWith(other: Observable<T>): Observable<T> = concat(this, other)
+
+fun <T> Observable<T>.concatWith(value: T): Observable<T> = concat(this, value.toObservable())
```
arkivanov: One more thing, to avoid clashing if we will add more overloads later, let's rename it to 'startWithValue', same way as in RxJava. Same for 'startWith'. Later we might add operators like 'startWithArray', 'startWithList' etc.

dmitryustimov: Sure, but won't it be better to use methods like this with named arguments? For example,
`.startWith(value = Event.Started)` or `.startWith(iterable = listOf(...))`. What kind of names are preferred?

arkivanov: I would stick with `startWithValue`-like names, from my point of view it is error prone to have same names. If there are two methods `startWith(T)` and `startWith(List<T>)` then you can by accident call without named argument:
```
val observable: Observable<List<String>> = ...
observable.startWith(emptyList())
```

I can't predict which method will be chosen. So better to be safe even without named arguments.

dmitryustimov: Done

PR #143, opened 2019-07-31
Fixed broken KN Condition again :-(

arkivanov:


arkivanov:
@CherryPerry Removed ConditionTest as it appears to impossible to test Condition properly. `await` may be interrupted at any time before timeout, it's documented actually.

arkivanov:
@CherryPerry Sorry, one more commit :-(

PR #145, opened 2019-08-01
Added onErrorReturn, onErrorResumeNext and onErrorComplete operators

dmitryustimov:


reaktive/src/commonMain/kotlin/com/badoo/reaktive/completable/OnErrorResumeNext.kt
```diff
@@ -0,0 +1,37 @@
...
+
+        subscribeSafe(
+            object : CompletableObserver, CompleteCallback by observer {
+
+                override fun onSubscribe(disposable: Disposable) {
+                    disposableWrapper.set(disposable)
+                }
+
+                override fun onError(error: Throwable) {
+                    observer.tryCatch({ nextSupplier(error) }) {
```
arkivanov: If `nextSupplier` throws an exception then there are two exceptions: `error` and the new one. You should compose them into `CompositeException`. Check out `doOnBeforeError` operator for usage example. Same for other similar places. Also please fix this for `Observable.onErrorResumeNext`, looks like I missed when I was reviewing that.

arkivanov:
We are going to release `1.0.0`, let's have it there :+1:

reaktive/src/commonMain/kotlin/com/badoo/reaktive/completable/OnErrorResumeNext.kt
dmitryustimov: Thanks for the reference to `doOnBeforeError`. Now both exceptions will be wrapped into `CompositeException` when supplier call fails.
Also please take a look at `TestObserver.assertError` with predicate

reaktive/src/commonTest/kotlin/com/badoo/reaktive/maybe/OnErrorResumeNextTest.kt
```diff
@@ -119,11 +120,16 @@ class OnErrorResumeNextTest :
...
-
-        observer.assertError(throwable)
+        val upstreamThrowable = Throwable()
+        val supplierThrowable = Throwable()
+        val observer = upstream.onErrorResumeNext { throw supplierThrowable }.test()
+        upstream.onError(upstreamThrowable)
+
+        observer.assertError { throwable ->
+            throwable is CompositeException &&
+                    throwable.cause1 == upstreamThrowable &&
```
arkivanov: We are following Kotlin code styles, continuation indent should be 4 spaces, not 8.
We will consider to fix it in the repo somehow, but please change it manually in IDEA settings and reformat all the code.
image

dmitryustimov: Fixed

PR #148, opened 2019-08-02
Added timer operators and tests

dmitryustimov:


reaktive/src/commonMain/kotlin/com/badoo/reaktive/maybe/Timer.kt
```diff
@@ -0,0 +1,10 @@
+package com.badoo.reaktive.maybe
+
+import com.badoo.reaktive.scheduler.Scheduler
+
+fun maybeTimer(delayMillis: Long, scheduler: Scheduler): Maybe<Long> =
```
arkivanov: Isn't it would be better to emit Unit? Same for other variants?

sample-android-app/src/main/java/com/badoo/reaktive/sample/android/MainActivity.kt
```diff
@@ -24,11 +24,8 @@ class MainActivity : AppCompatActivity() {
         val textView = findViewById<TextView>(R.id.text)
 
         findViewById<Button>(R.id.button).setOnClickListener {
-            observableFromFunction<String> {
-                Thread.sleep(1000L)
```
arkivanov: The idea was to simulate blocking operation on background thread here. Later I want to replace it with Kitties loading (like in JS sample).

reaktive-test/src/commonMain/kotlin/com/badoo/reaktive/test/scheduler/TestScheduler.kt
```diff
@@ -13,7 +13,7 @@ class TestScheduler(
 
     val timer = Timer()
     private val _executors: AtomicReference<List<Executor>> = AtomicReference(emptyList(), true)
-    val executors: List<Executor> = _executors.value
+    val executors: List<Executor> get() = _executors.value
```
arkivanov: :+1:

reaktive/src/commonMain/kotlin/com/badoo/reaktive/maybe/Timer.kt
dmitryustimov: It's a good question. It seems like other libraries (RxJava, RxJs, RxSwift) emits number. RxJava emits 0L, RxJs emits delay value. But Unit is also good choice.
I think delay value can be useful in complex data flows like this:
`
longObservable.flatMap { delay ->
    Observable.timer(delay, computationScheduler)
         .map(Long::toString)
         .startWith("Timer started")
}
`

dmitryustimov: Also I think about replacing `delayMillis: Long` by time span data type. I was looking at [TimeSpan](https://github.com/korlibs/klock/blob/master/klock/src/commonMain/kotlin/com/soywiz/klock/TimeSpan.kt).
Time span data type instead of Long can prevent some kind of errors. But we should pay attention to backward  compatibility.

arkivanov: Let's emit delay value. 0L is meaningless and useless. TimeSpan is a good thing but it's external dependency we don't have at the moment. We will consider it though. I'll create an issue.

sample-android-app/src/main/java/com/badoo/reaktive/sample/android/MainActivity.kt
dmitryustimov: Reverted

reaktive/src/commonMain/kotlin/com/badoo/reaktive/maybe/Timer.kt
dmitryustimov: Fixed. Now timer operator emits delay value.

PR #149, opened 2019-08-03
Added flatMap overloads with result selector

dmitryustimov:


reaktive/src/commonMain/kotlin/com/badoo/reaktive/maybe/FlatMap.kt
```diff
@@ -30,4 +30,11 @@ fun <T, R> Maybe<T>.flatMap(mapper: (T) -> Maybe<R>): Maybe<R> =
...
             }
         )
-    }
\ No newline at end of file
+    }
+
+fun <T, U, R> Maybe<T>.flatMap(
+    mapper: (T) -> Maybe<U>,
+    resultSelector: (T, U) -> R
+): Maybe<R> = flatMap { t ->
```
arkivanov: Please wrap line after `=` in case of multi-line function body. Also consider single for the method signature, it is less than 100 symbols. Same for other cases.

dmitryustimov: I'm not sure if I understood you correctly. Please review changes

reaktive/src/commonMain/kotlin/com/badoo/reaktive/single/FlatMapObservable.kt
```diff
@@ -2,6 +2,15 @@ package com.badoo.reaktive.single
...
 import com.badoo.reaktive.observable.Observable
 import com.badoo.reaktive.observable.flatMap
+import com.badoo.reaktive.observable.map
 
 fun <T, R> Single<T>.flatMapObservable(mapper: (T) -> Observable<R>): Observable<R> =
-    asObservable().flatMap(mapper)
\ No newline at end of file
+    asObservable().flatMap(mapper)
+
+fun <T, U, R> Single<T>.flatMapObservable(
```
arkivanov: Please choose one style, I would put whole definition on one line if it fits. Now some are multi line and some are single line. Max line width is 140

dmitryustimov: Fixed

PR #157, opened 2019-08-07
 Added Observable.debounce operator with selector. Added tests

dmitryustimov:


reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/Debounce.kt
```diff
@@ -1,66 +1,7 @@
...
-                }
-            }
-        )
-    }
-
-private class DebouncePendingValue<T>(
-    val value: T
-)
\ No newline at end of file
+    debounce { completableTimer(timeoutMillis, scheduler) }
```
arkivanov: This is much less efficient then it was as there is a Completable per every item and each creates an instance of Executor. Let's stick with two different implementations. Think about maybe it's possible to extract some common logic to a method or class.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/DebounceWithSelector.kt
```diff
@@ -0,0 +1,82 @@
...
+                                    serializedEmitter.onNext(value)
+                                }
+                            }
+                        }
+
+                    completable.subscribeSafe(innerObserver)
+                }
+
+                override fun onComplete() {
+                    val previousPendingValue = pendingValue.getAndUpdate { null }
```
arkivanov: The variable can be inlined

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/Debounce.kt
```diff
@@ -1,66 +1,7 @@
...
-                    }
-                }
-
-                override fun onComplete() {
-                    executor.cancel()
-
-                    executor.submit {
-                        pendingValue
-                            .value
-                            ?.value
```
arkivanov: This is actually bug, if `value` is null then `onNext` won't be called. Please add test for the case. For both variants.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/DebounceWithSelector.kt
```diff
@@ -0,0 +1,82 @@
...
+                    innerDisposableWrapper.set(localDisposableWrapper)
+
+                    val innerObserver =
+                        object : CompletableObserver, ErrorCallback by serializedEmitter {
+                            override fun onSubscribe(disposable: Disposable) {
+                                localDisposableWrapper.set(disposable)
+                            }
+
+                            override fun onComplete() {
+                                if (pendingValue.value === newPendingValue) {
```
arkivanov: Race condition, this `onComplete` may be executed concurrently with the main `onComplete`, please check the approach from the original source.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/Debounce.kt
dmitryustimov: It seems that only `DebouncePendingValue` can be shared between two implementations because other common parts are too small to be extracted. So I made `DebouncePendingValue` internal instead of private.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/Debounce.kt
dmitryustimov: Should be fixed, please review

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/DebounceWithSelector.kt
dmitryustimov: Should be fixed, please review

PR #158, opened 2019-08-09
Added Observable.toMap operator

dmitryustimov:


reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/ToMap.kt
```diff
@@ -0,0 +1,13 @@
...
+import com.badoo.reaktive.single.Single
+
+fun <T, K> Observable<T>.toMap(keySelector: (T) -> K): Single<Map<K, T>> =
+    toMap(keySelector) { it }
+
+fun <T, K, V> Observable<T>.toMap(keySelector: (T) -> K, valueSelector: (T) -> V): Single<Map<K, V>> =
+    collect(emptyMap()) { map, item ->
+        val key = keySelector(item)
+        val value = valueSelector(item)
+        return@collect map + mapOf(key to value)
```
CherryPerry: You can use `return@collect map + (key to value)` to avoid allocation of map.
And I am not sure, but probably we can remove `return@collect` because last expression's result will be returned automatically from lambda.

dmitryustimov: I've seen both variants of using or not using `return@lambda`. Which one is preferred?

arkivanov: Yeah, two things here:

1. `map + mapOf(k to v)` can be replaced with `map + (k to v)` or just `map.plus(k to v)`
2. `return@collect` should be omitted, it makes sense when you need to return from the middle of lamda, or from some outer lambda.

PR #172, opened 2019-08-24
iOS sample should use sample-mpp-module

CherryPerry:
Fixes https://github.com/badoo/Reaktive/issues/164

Simulator Screen Shot - iPhone Xʀ - 2019-08-24 at 17 44 07


sample-ios-app/xcode/sample-ios-app/KittenStoreImpl.swift
```diff
@@ -0,0 +1,25 @@
+import Foundation
+import ReactiveSample
+import Alamofire
+
+class KittenLoaderImpl: KittenLoader {
```
arkivanov: So you want to use some platform specific things for networking. I really want to avoid using Reaktive outside of Kotiln. KittenStoreBuilder is designed to accept platform specific dependencies. What do you think about the following approach?

1. Create an interface in `sample-mpp-module/iosCommonMain` like:
```
interface KittenDataSource {   

    fun load(url: String): String
}
```

2. Implement this interface on iOS side.

3. Implement `KittenLoaderImpl` like:
```
internal class KittenLoaderImpl(
    private val dataSource: KittenDataSource
) : KittenLoader {

    override fun load(): Single<Result> =
        singleFromFunction { dataSource.load(Config.KITTEN_URL) }
            .subscribeOn(ioScheduler)
            .map { ... }
            .map(Result::Success)
            .onErrorReturnValue(Result.Error)
}
```

4. Implement `KittenStoreBuilderImpl` like:
```
class KittenStoreBuilderImpl(
    private val dataSource: KittenDataSource
) : KittenStoreBuilder {

    override fun build(): KittenStore = KittenStoreImpl(loader = KittenLoaderImpl(dataSource = dataSource))
}
```

sample-ios-app/src/iosCommonMain/kotlin/com.badoo.reaktive.sample.ios/ReaktiveSample.kt
```diff
@@ -0,0 +1,6 @@
+package com.badoo.reaktive.sample.ios
```
arkivanov: Just noticed, you have wrong folder name. The name is "com.badoo.reaktive.sample.ios". Folders should be nested instead.

sample-mpp-module/src/iosCommonMain/kotlin/com/badoo/reaktive/samplemppmodule/KittenDataSource.kt
```diff
@@ -0,0 +1,6 @@
+package com.badoo.reaktive.samplemppmodule
+
+interface KittenDataSource {
+
+    fun load(url: String, continuation: (String?, Throwable?) -> Unit)
```
arkivanov: I'm not sure, but maybe response should be a sealed class with two options.

PR #173, opened 2019-08-25
Implemented thread local subscribe

arkivanov:
This is about freezing objects in K/N. Normally the following code will cause freezing of any objects captured by lambdas, e.g. `list` will be freezed:
```
fun foo(source: Observable<Int>) {
    val list = ArrayList<Int>()

    source
        .observeOn(mainScheduler)
        .subscribe { list.add(it) }
}
```

However if the `foo` is called on main thread and we are observing values on main thread (or any other same thread), then it would be good to avoid freezing. This PR introduces possibility to subscribe without freezing, but only for cases if threads are same.

Example:
```
fun foo(source: Observable<Int>) {

    val list = ArrayList<Int>()

    source
        .observeOn(mainScheduler)
        .subscribe(isThreadLocal = true) { list.add(it) }
}
```

`list` won't be freezed, but can be updated only from same thread.


reaktive/src/commonMain/kotlin/com/badoo/reaktive/utils/threadlocal/ThreadLocalState.kt
```diff
@@ -0,0 +1,22 @@
+package com.badoo.reaktive.utils.threadlocal
+
+import kotlin.native.concurrent.ThreadLocal
+
+@ThreadLocal
```
CherryPerry: It is quite dangerous.

What if I make a mistake and call observable callback function on different thread? My callback won't be invoked and I won't get any error. Even warning. Quite easy to miss this mistake. AFAIK, invoking `dispose` does not guarantee, that callback functions won't be invoked because of multi-threading and synchronization issues. So let's do the next thing: we will throw `IllegalStateException`, when storage is initialized, `clear` was not invoked, but getter of `value` returns `null`. Doing this we can alert our users, that they are doing something wrong. They will able even report this issue to any crash tracking system. We won't cover issue, when callback is called after `dispose` on different thread, but if you already unsubscribed, who cares.

What if I make a mistake and dispose my thread local subscription on different thread? Looks like it will become memory leak. Following changes from above, we can throw `IllegalStateException` when you try to invoke `clear` with no value inside.

Also looks like throwing this exception should not be passed to `onError` handler, it should be thrown directly to global error handler. If it is possible to add some information about threads in this exception, let's do it! For example creator thread's id and name and invoker thread's id and name.

CherryPerry: > I believe that such errors should be delivered via normal `onError` flow
Yes, you are right. In both cases `onError` callback will be `null`. That means error would be propagated to `reaktiveUncaughtErrorHandler`.

CherryPerry: 1. When I receive callback not on subscription creator's thread
2. When I dispose subscription not on subscription creator's thread

In both cases I won't have any `onError` delegate, because I try to get it on different thread.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/utils/ThreadLocalStorage.kt
```diff
@@ -0,0 +1,78 @@
...
+        checkCurrentThread()
+
+        if (_isDisposed.compareAndSet(false, true)) {
+            ThreadLocalState[key] = null
+        }
+    }
+
+    private fun checkCurrentThread() {
+        if (currentThreadId != initialThreadId) {
+            throw RuntimeException("Accessing ThreadLocalStorage from another threads is prohibited")
```
CherryPerry: Add thread's info to exception's message.

reaktive/src/jvmNativeCommonTest/kotlin/com/badoo/reaktive/utils/ThreadLocalStorageThreadingTest.kt
```diff
@@ -0,0 +1,59 @@
+package com.badoo.reaktive.utils
+
+import com.badoo.reaktive.scheduler.computationScheduler
+import com.badoo.reaktive.utils.atomic.AtomicReference
+import kotlin.test.Test
+import kotlin.test.assertTrue
+
+class ThreadLocalStorageThreadingTest {
```
CherryPerry: Merge with `ThreadLocalStorageTest`. They previously were testing different classes, but now it is only one.

reaktive/src/jvmNativeCommonTest/kotlin/com/badoo/reaktive/utils/ThreadLocalStorageThreadingTest.kt
```diff
@@ -0,0 +1,59 @@
...
+
+            lock.synchronized {
+                isFailedRef.value = isFailed
+                condition.signal()
+            }
+        }
+
+        lock.synchronized {
+            while (isFailedRef.value == null) {
+                condition.await()
```
CherryPerry: `await` with timeout to not deadlock CI.

PR #177, opened 2019-08-29
Handle exceptions in subscribe callbacks

### LINKED ISSUE #176: Exceptions in subscribe callbacks are not handled
Currently all exceptions thrown by `subscribe` callbacks (like `onSubscribe`, `onNext`, `onComplete`, etc.) are handled by neither `onError` nor `uncaughtExceptionHandler`. `onError` callback should be user for non-terminal events like `onSubscribe` or `onNext` with fallback to `uncaughtExceptionHandler`. And `uncaughtExceptionHandler` should be used directly for terminal events like `onComplete`, `onSuccess` or `onError`.

arkivanov:
Quite a big change, extracted boilerplate to a separate file.
Fixes #176

### DIFF
--- a/reaktive/src/commonMain/kotlin/com/badoo/reaktive/utils/PrintError.kt
+++ b/reaktive/src/commonMain/kotlin/com/badoo/reaktive/utils/PrintError.kt
@@ -1,3 +1,14 @@
 package com.badoo.reaktive.utils
 
+import com.badoo.reaktive.utils.atomic.AtomicBoolean
+
+@Suppress("ObjectPropertyName") // Backing field
+private val _isPrintErrorEnabled = AtomicBoolean(true)
+
+internal var isPrintErrorEnabled
+    get() = _isPrintErrorEnabled.value
+    set(value) {
+        _isPrintErrorEnabled.value = value
+    }
+
 internal expect fun printError(error: Any?)
--- a/reaktive/src/commonMain/kotlin/com/badoo/reaktive/utils/Utils.kt
+++ b/reaktive/src/commonMain/kotlin/com/badoo/reaktive/utils/Utils.kt
@@ -1,5 +1,7 @@
 package com.badoo.reaktive.utils
 
+import com.badoo.reaktive.base.exceptions.CompositeException
+
 internal fun handleSourceError(error: Throwable, onError: ((Throwable) -> Unit)? = null) {
     try {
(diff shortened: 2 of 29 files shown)


reaktive/src/commonMain/kotlin/com/badoo/reaktive/utils/Utils.kt
```diff
@@ -10,7 +12,8 @@ internal fun handleSourceError(error: Throwable, onError: ((Throwable) -> Unit)?
             } catch (e: Throwable) {
```
CherryPerry: Please, rename `e` into `errorHandlerException` or whatever. It is hard to distinct them.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/utils/PrintError.kt
```diff
@@ -1,3 +1,14 @@
 package com.badoo.reaktive.utils
 
+import com.badoo.reaktive.utils.atomic.AtomicBoolean
+
+@Suppress("ObjectPropertyName") // Backing field
+private val _isPrintErrorEnabled = AtomicBoolean(true)
+
+internal var isPrintErrorEnabled
```
CherryPerry: Use it for `printStack()` too. It is quite strange, that you can skip some logging of small messages like `Some shit happened` but not huge stacktraces.

Maybe in future we should introduce `Logger` that can log text and exceptions (with stacktraces) and default implementation of it in the way it is implemented now.
For example on Android it will become possible to use `Tinder` to log it and report to server.

CherryPerry: Let's keep it as is for now, but introduce `Logger ` interface later, so we can mock logger for android.

arkivanov:
@CherryPerry Added more tests, found and fixed a few bugs.

arkivanov:
iOS build failed as always. Will fix using logs..............

arkivanov:
Added tests for `currentThreadId` to make sure it returns same values for same threads and different values for different threads.

arkivanov:
Also moved `doInBackground` from `commonTest` to `jvmNativeCommonTest` and split for Native and JVM. Because scheduler based approach reuses threads but for tests we need new threads every time.

PR #180, opened 2019-09-01
Fix maybeOf nullability

arkivanov:
Perhaps `maybeOf(value)` should work as in RxJava2, if value is `null` then `Maybe` should just complete, otherwise emit the value. Not sure though.


arkivanov:
Don't like it

PR #184, opened 2019-09-03
Added threadLocal operator

arkivanov:
Currently we have `isThreadLocal` flag that can be used like `.subscribe(isThreadLocal = true, ...)`. This allows to subscribe to an upstream that produces events on the same thread as subscription without freezing `subscribe`'s callbacks.

We can do better by introducing `threadLocal()` operator. This will allow us to avoid freezing in cases like:
```
val list = ArrayList<Int>() // List will not be frozen!

observableOf(1)
    .observeOn(mainScheduler)
    .threadLocal()
    .map { ... } // Callback will not be frozen!
    .doOnBeforeNext { list.add(1) } // Callback will not be frozen!
    .subscribe() // Must happen on Main thread!
```

We still have `isThreadLocal` flag in `subscribe` which is very useful, it just uses `threadLocal()` internally now.


arkivanov:
> Have a few questions about the code review:
> In ThreadLocalStorage we have an AtomicBoolean for a field that will only be accessed on a single thread due to the checkCurrentThread calls.
> Is this because we expect ThreadLocalStorage to be frozen at some point?
> 
> So the idea is that if the lambdas get frozen, all their arguments will also get frozen and they can't make calls like "list.add(xyz)". By storing the callback in a ThreadLocal we avoid them being frozen allowing the lambdas to perform mutation (instead the ThreadLocalStorage gets frozen, and it doesn't have a ref to the callback so they don't get frozen). Is that correct?

Yes this is correct. Instances of `ThreadLocalStorage` are shared between threads so they are frozen. That's why we need `AtomicBoolean` under the hood. But because storages' values are stored in the `@ThreadLocal ThreadLocalState`, they are not frozen but can be retrieved only from the original thread.

PR #185, opened 2019-09-04
Make publication name match MPP naming model

CherryPerry:
It is much easier just to rename publication name (that is used only in Gradle) than to do fixes inside plugin.


arkivanov:
@CherryPerry If it won't affect anything else than publication then I'm fine

CherryPerry:
Should not affect. At least now it is published properly.

Screenshot 2019-09-04 at 22 32 25

PR #188, opened 2019-09-04
Use FreezableAtomicReference

arkivanov:
Well, looks like it is actually a good idea to use FreezableAtomicReference. It allows us to freeze things on demand, when needed.

Examples:
```
        observable<Int> { emitter ->
            val myDisposable: Disposable = ...
            emitter.setDisposable(myDisposable) // Will not be frozen
        }
            .subscribe()
```

```
        observable<Int> { emitter ->
            lateinit var myDisposable: Disposable
            emitter.setDisposable(myDisposable) // Will be frozen
        }
            .subscribe()
```

There 3 commits:
1. Introducing FreezableAtomicReference
2. Stupid code refactoring (big one, can be skipped if not interested)
3. Some threading tests

Only one thing worries me: FreezableAtomicReference was introduced in Kotlin 1.3.40 as a no-priority feature and in its [release notes](https://blog.jetbrains.com/kotlin/2019/06/kotlin-1-3-40-released/) called as experimental prototype. However it is not annotated as experimental and the [sources](https://github.com/JetBrains/kotlin-native/blob/66b813a6a288cc2a0f0e25e751e5c1d555c78f24/runtime/src/main/kotlin/kotlin/native/concurrent/Atomics.kt#L282) look pretty fine (checked multiple times).

It will also help us a lot with coroutines interop (working on it)


arkivanov:
@CherryPerry please read the ending of the PR description. What do you think of using FreezableAtomicReference?

reaktive/src/jvmNativeCommonTest/kotlin/com/badoo/reaktive/utils/atomic/AtomicReferenceThreadingTest.kt
```diff
@@ -0,0 +1,55 @@
+package com.badoo.reaktive.utils.atomic
+
+import com.badoo.reaktive.test.doInBackgroundBlocking
+import kotlin.test.Test
+import kotlin.test.assertFalse
+import kotlin.test.assertSame
+import kotlin.test.assertTrue
+
+class AtomicReferenceThreadingTest {
```
CherryPerry: Add test for freeze too? Verify that in some cases referenced value is frozen and in other is not.

PR #193, opened 2019-09-13
Add Observable.take(Int) operator

saket:
I added a bunch of tests along the lines of how you've structured other tests. Let me know if you want me to add more or modify the tests. Please feel free to make the changes yourself as well if that's convenient/faster for you.

There was one test that I couldn't make pass unfortunately so I'll need help here: 

```kotlin
@Test
fun disposes_WHEN_upstream_disposed() {
    upstream.dispose()

    observer.assertDisposed()
}
```


reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/TakeObservable.kt
```diff
@@ -0,0 +1,43 @@
...
+
+import com.badoo.reaktive.base.subscribeSafe
+import com.badoo.reaktive.disposable.Disposable
+import com.badoo.reaktive.disposable.DisposableWrapper
+
+/**
+ * Emit only the first [limit] items emitted by source.
+ */
+fun <T> Observable<T>.take(limit: Int): Observable<T?> {
+    require(limit > 0) { "count >= 0 required but it was $limit" }
```
CherryPerry: count **>** 0 required but it was $limit

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/TakeObservable.kt
```diff
@@ -0,0 +1,43 @@
...
+
+    return observable { emitter ->
+        val disposableWrapper = DisposableWrapper()
+        emitter.setDisposable(disposableWrapper)
+
+        var remaining = limit
```
CherryPerry: `val remaining = AtomicInt(limit)`

Use atomics to support KN memory model and restrictions.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/TakeObservable.kt
```diff
@@ -0,0 +1,43 @@
...
+
+        subscribeSafe(object : ObservableObserver<T?> {
```
CherryPerry:
```kotlin
object : ObservableObserver<T>, ErrorCallback by emitter, CompleteCallback by emitter {
    fun onNext()
    fun onSubscribe()
}
```

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/TakeObservable.kt
```diff
@@ -0,0 +1,43 @@
...
+            override fun onComplete() {
+                emitter.onComplete()
+            }
+
+            override fun onError(error: Throwable) {
+                emitter.onError(error)
+            }
+
+            override fun onNext(value: T?) {
+                if (remaining-- > 0) {
```
CherryPerry: @arkivanov As I remember we don't need any synchronisation here, right?

reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/TakeObservableTests.kt
```diff
@@ -0,0 +1,109 @@
...
+            .test()
+
+        upstream.onNext(42)
+
+        observer.assertValue(42)
+        observer.assertComplete()
+    }
+
+    @Test fun throws_error_when_take_is_called_with_a_limit_of_0() {
+        assertFailsWith<IllegalArgumentException>("count >= 0 required but it was 0") {
```
CherryPerry: `count > 0`

And I am not sure about checking exception message, just `IllegalArgumentException` should be fine.

reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/TakeObservableTests.kt
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    @Test
+    fun multiple_takes() {
+        val observer = upstream
+            .take(100)
+            .take(1)
+            .test()
+
+        (1..100).forEach {
```
CherryPerry: I think `upstream.onNext(1, 2)` is enough.

reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/TakeObservableTests.kt
```diff
@@ -0,0 +1,109 @@
...
+    fun produces_same_error_WHEN_upstream_produced_error() {
+        val error = Exception()
+
+        upstream.onError(error)
+
+        observer.assertError(error)
+    }
+
+    @Test
+    fun disposes_WHEN_upstream_disposed() {
```
CherryPerry: @arkivanov Please check this test, don't see any problems in code.

reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/TakeObservableTests.kt
```diff
@@ -0,0 +1,109 @@
...
+import com.badoo.reaktive.test.base.assertNotError
+import com.badoo.reaktive.test.observable.*
+import kotlin.test.Test
+import kotlin.test.assertFailsWith
+
+class TakeObservableTests :
+    ObservableToObservableTests by ObservableToObservableTests<Unit>({ take(1) }) {
+
+    private val upstream = TestObservable<Int?>()
+    val observer = upstream.take(10).test()
```
CherryPerry: `private`?

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/TakeObservable.kt
arkivanov: I would say 0 should be a valid value, so `require(limit >= 0)`.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/TakeObservable.kt
arkivanov: Just `CompletableCallbacks by emitter` will work as well. `CompletableCallbacks` extends both `ErrorCallback` and `CompleteCallback`.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/TakeObservable.kt
arkivanov: All `Observer`'s callbacks are never executed concurrently, but may be executed from different threads. As we are using atomics then no syncrhonization is needed.

reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/TakeObservableTests.kt
arkivanov: The test is invalid. You should not dispose upstream manually, operators are responsible for it.
In this case `Disposable` passed to downstream `Observer` should be disposed when: 

1. Upstream produced a terminal event (`Complete` or `Error`)
2. Upstream produced a required amount of values

The first case should be covered in basic tests (`ObservableToObservableTests` etc.), it is not now, filed issue #195.
The second case should be covered here, like `disposes_disposable_WHEN_values_limit_is_reached`.

reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/TakeObservableTests.kt
```diff
@@ -0,0 +1,109 @@
...
+
+    @Test
+    fun produces_values_in_correct_order_from_upstream() {
+        upstream.onNext(42, null, 84, null)
+
+        observer.assertValues(42, null, 84, null)
+    }
+
+    @Test
+    fun multiple_takes() {
```
arkivanov: The test is useless, it's enough to test just one `take` and use `TestObservable` as upstream.

reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/TakeObservableTests.kt
arkivanov: 0 should be a valid value. Please test for failure on `-1`.

reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/TakeObservableTests.kt
```diff
@@ -0,0 +1,109 @@
...
+        (1..100).forEach {
+            upstream.onNext(it)
+        }
+
+        observer.assertValue(1)
+        observer.assertComplete()
+    }
+
+    @Test
+    fun re_entrant_take() {
```
arkivanov: "reentrant"?

reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/TakeObservableTests.kt
```diff
@@ -0,0 +1,109 @@
...
+        val observer = upstream
+            .take(1)
+            .doOnBeforeNext {
```
arkivanov: Better to avoid using real operators in tests as they may affect the test. Just use `let' and a simple anonymous `Observable`.

reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/TakeObservableTests.kt
arkivanov: Please specify at least the expected behavior in names. Like `counts_recursive_invocations` or similar.

reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/TakeObservableTests.kt
```diff
@@ -0,0 +1,109 @@
...
+                upstream.onNext(84)
+            }
+            .test()
+
+        upstream.onNext(42)
+
+        observer.assertValue(42)
+        observer.assertComplete()
```
arkivanov: This assertion is out of scope of the test.

reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/TakeObservableTests.kt
```diff
@@ -0,0 +1,109 @@
...
+    val observer = upstream.take(10).test()
+
+    @Test
+    fun completes_WHEN_limit_of_1_is_reached() {
+        val observer = upstream.take(1).test()
+
+        upstream.onNext(42)
+
+        observer.assertComplete()
+        observer.assertNotError()
```
arkivanov: Useless assertion, it is verified by TestObserver.

reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/TakeObservableTests.kt
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    @Test
+    fun completes_WHEN_limit_of_2_is_reached() {
+        val observer = upstream.take(2).test()
+
+        upstream.onNext(42)
+        observer.assertNotComplete()
```
arkivanov: Assertion is out of scope of the test, as you are verifying "completes when limit of 2 is reached". So produce two values and verify that completed.

arkivanov: And add another test to verify that "does not complete when limit is not reached".

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/TakeObservable.kt
saket: Can you think of a valid usecase for allowing `take(0)`? Isn't that as good as completing/disposing upstream right away? 

I actually copied this check from RxJava.

reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/TakeObservableTests.kt
saket: Updated the error message. Continuing this conversation here: https://github.com/badoo/Reaktive/pull/193#discussion_r324485860

reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/TakeObservableTests.kt
saket: Just to confirm, you're suggesting to remove this test right?

reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/TakeObservableTests.kt
saket: I actually couldn't think of a good name. Wanna help out? 

The usecase I'm trying to test here is when `upstream` produces a value from inside `onNext()`.

reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/TakeObservableTests.kt
saket: Good idea.

reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/TakeObservableTests.kt
saket: I'm not sure what you mean by `Just use let' and a simple anonymous`. Can you share some code?

reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/TakeObservableTests.kt
saket: > The first case should be covered in basic tests (ObservableToObservableTests etc.), it is not now, filed issue #195.

Sounds good!

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/TakeObservable.kt
arkivanov: I could imagine one, if limit is calculated somehow or received from e.g. server, like:
`users.take(server.getUserLimit()).otherOperators()`
Here 0 might be a valid value, so you don't need to care about it. If 0 would not be allowed, then you will have to write something like:
```
users
    .let { source ->
        server
            .getUserLimit()
            .takeIf { it > 0 }
            ?.let(source::limit) 
            ?: source
    }
    .otherOperators()
```

Also 0 is allowed in stdlib functions like `Iterable<T>.take(n)`. Also RxJava2's `Observable.take` allows 0.

reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/TakeObservableTests.kt
arkivanov: Yes you can remove it, based on the assumption that you tested this behaviour using `TestObservable`.

reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/TakeObservableTests.kt
arkivanov: I think something like this should work?
```
        upstream
            .take(1)
            .let { source ->
                observableUnsafe<Int?> { observer ->
                    source.subscribe(
                        object : ObservableObserver<Int?> by observer {
                            override fun onNext(value: Int?) {
                                upstream.onNext(48)
                                observer.onNext(value)
                            }
                        }
                    )
                }
            }
            .test()
```

reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/TakeObservableTests.kt
arkivanov: So for this one I think 'counts_recursive_invocations' would be more meaningful.

saket:
@arkivanov all changes done. Please take a look?

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/TakeObservable.kt
```diff
@@ -0,0 +1,41 @@
...
+        val disposableWrapper = DisposableWrapper()
+        emitter.setDisposable(disposableWrapper)
+
+        val remaining = AtomicInt(limit)
+
+        subscribeSafe(object : ObservableObserver<T>, CompletableCallbacks by emitter {
+            override fun onSubscribe(disposable: Disposable) {
+                disposableWrapper.set(disposable)
+
+                if (remaining.value == 0) {
```
arkivanov: Don't you think that if limit is 0 then there is no need to subscribe to the upstream at all? Perhaps you can just call `emitter.onComplete()` and return right after `emitter.setDisposable(...)` in this case. RxJava2 actually subscribes and then immediately completes from `onSubscribe`, exactly as you did. But I don't see any points of such a behaviour. Better to just do nothing.

arkivanov: Please also make sure there is a test for either case: explicit test that it does not subscribe to the upstream if `limit` is 0.

saket: > Don't you think that if limit is 0 then there is no need to subscribe to the upstream at all?

I believe there is an important distinction: the disposal only happens once it's subscribed. I think that has always been the expected behavior for Rx where cogs start moving only when they're subscribed and we shouldn't be making Reaktive incompatible.

saket: > Please also make sure there is a test for either case: explicit test that it does not subscribe to the upstream if limit is 0.

This one?
https://github.com/badoo/Reaktive/pull/193/files#diff-00df270188cf11ffd754c95a7e37d345R15

arkivanov: You are getting me wrong. If you will not subscribe to the upstream and just complete then you will not violate any rx rules. You don't even need to set a desposable for the emitter. Just call emitter.onComplete() and return from the very beginning.

arkivanov: So if you write something like:
```
    observable { emitter ->
        if (limit == 0) {
            emitter.onComplete()
            return@observable
        }
        
        // rest of the code
```

then it should be fine. Downstream observer`s `onSubscribe` will be already called at this point by `observable{}` builder function and you are free to produce any events from the beginning. There is no obligation to subscribe to the upstream.

arkivanov: Alright, looks like it is better to follow RxJava in this case.

arkivanov:
@CherryPerry please review the changes

arkivanov:
@CherryPerry the PR stuck waiting for status, there are no running tasks. Do you have any idea of how we can fix it?

Issue #195, opened 2019-09-15
Some operators are not disposing their Disposables on terminal events


arkivanov:
We should extend our generic tests (like `ObservableToObservableTests`) and add the following tests:

- `disposes_disposable_WHEN_upstream_completed`
- `disposes_disposable_WHEN_upstream_succeeded`
- `disposes_disposable_WHEN_upstream_produced_error`

PR #197, opened 2019-09-15
Add Retry operator

CherryPerry:


reaktive/src/commonMain/kotlin/com/badoo/reaktive/completable/Retry.kt
```diff
@@ -0,0 +1,47 @@
...
+            object : CompletableObserver, CompleteCallback by observer {
+
+                override fun onSubscribe(disposable: Disposable) {
+                    disposableWrapper.set(disposable)
+                }
+
+                override fun onError(error: Throwable) {
+                    observer.tryCatch(
+                        { predicate(attempt.addAndGet(1), error) },
+                        { CompositeException(error, it) }) { shouldRetry ->
```
arkivanov: Wrong formatting, ") {" should be on the new line.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/completable/OnErrorResumeNext.kt
```diff
@@ -20,7 +20,9 @@ fun Completable.onErrorResumeNext(nextSupplier: (Throwable) -> Completable): Com
                 }
 
                 override fun onError(error: Throwable) {
-                    observer.tryCatch({ nextSupplier(error) }, { CompositeException(error, it) }) {
+                    observer.tryCatch(
+                        { nextSupplier(error) },
+                        { CompositeException(error, it) }) {
```
arkivanov: Wrong formatting, ") {" should be on the new line.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/completable/Retry.kt
```diff
@@ -0,0 +1,47 @@
...
+                        if (shouldRetry) {
+                            // Should dispose previous before,
+                            // otherwise new subscription will be disposed via onSubscribe disposal
+                            disposableWrapper.set(null)
```
arkivanov: We should not really care about upstream's Disposable since it's already finished.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/completable/Retry.kt
```diff
@@ -0,0 +1,47 @@
...
+import com.badoo.reaktive.base.exceptions.CompositeException
+import com.badoo.reaktive.base.subscribeSafe
+import com.badoo.reaktive.base.tryCatch
+import com.badoo.reaktive.disposable.Disposable
+import com.badoo.reaktive.disposable.DisposableWrapper
+import com.badoo.reaktive.utils.atomic.AtomicInt
+import kotlin.reflect.KClass
+
+fun Completable.retry(predicate: (Int, Throwable) -> Boolean = { _, _ -> true }): Completable =
+    completableUnsafe { observer ->
```
arkivanov: Looks like we should prefer safe builder with emitter because it disposes an assigned Disposable on any terminal event. Otherwise your DisposableWrapper won't be disposed on e.g. completion. See #195.

arkivanov: Forget about it, I'm not sure at the moment. Will compare with rxjava later on. But perhaps you should still use emitter to avoid any events after 'onError', e.g. the predicate thrown an exception, you called 'onError' and then upstream still might produce an event.

CherryPerry: > the predicate thrown an exception, you called 'onError' and then upstream still might produce an event

But upstream should not produce any events after it produced error. When we in `onError` block, upstream is in terminated state.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/completable/Retry.kt
CherryPerry: Looks like it works fine now after `TestSource` changes.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/completable/Retry.kt
arkivanov: But in case when predicate throws exception you just calling `observer.onError`, so:
1. upstream is not disposed
2. even if you will dispose its disposable it still might emit a value and you have no protection against it

arkivanov: And the following test as well: `produces_error_WHEN_predicate_thrown_exception`

CherryPerry:
@arkivanov Review again please.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/completable/Retry.kt
```diff
@@ -0,0 +1,34 @@
...
+                    retry.onError(error) { this@retry.subscribeSafe(this) }
+                }
+            }
+        )
+    }
+
+fun Completable.retry(times: Int): Completable =
+    retry { attempt, _ -> attempt < times }
+
+fun Completable.retry(throwableType: KClass<out Throwable>): Completable =
```
arkivanov: Maybe also add inline reified variant?

reaktive/src/commonMain/kotlin/com/badoo/reaktive/completable/Retry.kt
```diff
@@ -0,0 +1,34 @@
+package com.badoo.reaktive.completable
+
+import com.badoo.reaktive.base.operator.Retry
+import com.badoo.reaktive.base.subscribeSafe
+import com.badoo.reaktive.disposable.Disposable
+import com.badoo.reaktive.disposable.DisposableWrapper
+import kotlin.reflect.KClass
+
+fun Completable.retry(predicate: (Int, Throwable) -> Boolean = { _, _ -> true }): Completable =
```
arkivanov: Let's add a name for `Int` parameter. Like `predicate: (attempt: Int, Throwable) -> ...`

reaktive/src/commonMain/kotlin/com/badoo/reaktive/completable/Retry.kt
```diff
@@ -0,0 +1,34 @@
...
+    completable { emitter ->
+        val retry = Retry(emitter, predicate)
```
arkivanov: Maybe better to put it inside observer anonymous class

reaktive/src/commonMain/kotlin/com/badoo/reaktive/completable/Retry.kt
CherryPerry: Screenshot 2019-09-22 at 20 47 08

CherryPerry: It clashes with base implementation.

Explicit statement below won't help too.
```kotlin
fun <T> Single<T>.retry(): Single<T> =
    retry { _, _ -> true }
```

arkivanov: You have to specify the type explicitly if there are more operators chained after `retry()`. This is fine. And I don't understand what exactly clashes. All the operators have different signatures.

CherryPerry: No, clash between `retry<Throwable>()` and `retry()` of `retry( {Int, Throwable} -> Boolean = { _, _ -> true} )`. Second one has default parameter and can be called like `retry(), that us why clash occurs.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/completable/Retry.kt
```diff
@@ -0,0 +1,33 @@
...
+                }
+            }
+        )
+    }
+
+fun Completable.retry(times: Int): Completable =
+    retry { attempt, _ -> attempt < times }
+
+fun Completable.retry(throwableType: KClass<out Throwable>): Completable =
+    retry { _, throwable -> throwableType.isInstance(throwable) }
```
arkivanov: Isn't `isInstance` requires reflection that is not available on Native?

CherryPerry: Covered with tests. iOS looks fine, CI will check Linux.

arkivanov: I have checked, it works on Linux. I'm still thinking about inline reified variant, it would be very nice to have.

I see two options:
1. remove default value for `predicate` argument
2. rename the last variant to `retryOnException` or similar

I would better go with the first option

PR #199, opened 2019-09-16
User synchronized instead of Lock in SerializerImpl in JVM and JS

arkivanov:
This PR addresses performance of `Serializer` and `BufferedExecutor` in JVM. As per JMH [benchmarks](https://github.com/ShikaSD/multiK/tree/master/bench/src) some operators are slower than RxJava2 (e.g. `flatMap`). This PR improves performance of `Serializer` and `BufferedExecutor` in JVM, so e.g. `flatMap` is now around ~3x faster.

### DIFF
--- a/reaktive/src/jvmJsCommonMain/kotlin/com/badoo/reaktive/utils/serializer/SerializerImpl.kt
+++ b/reaktive/src/jvmJsCommonMain/kotlin/com/badoo/reaktive/utils/serializer/SerializerImpl.kt
@@ -1,54 +1,68 @@
 package com.badoo.reaktive.utils.serializer
 
-import com.badoo.reaktive.utils.Lock
 import com.badoo.reaktive.utils.queue.Queue
-import com.badoo.reaktive.utils.queue.isNotEmpty
-import com.badoo.reaktive.utils.queue.take
-import com.badoo.reaktive.utils.synchronized
+import com.badoo.reaktive.utils.queue.isEmpty
 
 internal abstract class SerializerImpl<in T>(queue: Queue<T>) : Serializer<T> {
 
     private var queue: Queue<T>? = queue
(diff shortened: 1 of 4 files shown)


reaktive/src/jvmJsCommonMain/kotlin/com/badoo/reaktive/utils/serializer/SerializerImpl.kt
```diff
@@ -1,54 +1,68 @@
...
     private var isDraining = false
 
     override fun accept(value: T) {
-        lock
-            .synchronized {
-                queue
-                    ?.apply { offer(value) }
-                    ?.takeUnless { isDraining }
-                    ?.also { isDraining = true }
+        val q =
```
CherryPerry: Use `queue` and `this.queue`? I don't think `q` is good name for variable.

reaktive/src/jvmJsCommonMain/kotlin/com/badoo/reaktive/utils/serializer/SerializerImpl.kt
```diff
@@ -1,54 +1,68 @@
...
-                        take()
-                    } else {
+            val value =
+                synchronized(monitor) {
+                    if (isEmpty) {
                         onDrainFinished(false)
                         return
                     }
+
+                    poll()!!
```
CherryPerry: Why `poll()!!` instead of `take()`?

arkivanov: Because it checks isEmpty under the hood and I would like to avoid double check.
Discovered NPE and fixed.
Removed `take` as not being used.

PR #206, opened 2019-09-25
Introduce SharedList

arkivanov:
There are cases when using of `AtomicList` is inefficient in JVM and JS. The most common case is when access to the list is already synchronized by e.g. `Serializer` or by nature of the `Observer` callbacks. In this PR I'm introducing internal `SharedList` that implements `MutableList`. In JVM and JS it's backed by `ArrayList` and in Native it uses `AtomicList`, so it still can be frozen.

PS: replaced all usages of `AtomicList` with `SharedList` where possible
PPS: also made `AtomicList` and all its extensions internal - it's just typealias and we will have more control over it.
PPPS: `SharedList` will be very useful in next my PR about `Observable.replay` operator
PPPPS: There are 3 commits: `SharedList` itself, tests and refactoring

### DIFF
--- a/reaktive/src/commonTest/kotlin/com/badoo/reaktive/utils/SharedListIteratorTest.kt
+++ b/reaktive/src/commonTest/kotlin/com/badoo/reaktive/utils/SharedListIteratorTest.kt
@@ -1,0 +1,61 @@
+package com.badoo.reaktive.utils
+
+import kotlin.test.Test
+import kotlin.test.assertEquals
+
+class SharedListIteratorTest {
+
+    @Test
+    fun listIterator() {
+        val lists: Pair<MutableList<Int?>, MutableList<Int?>> = Pair(SharedList(), ArrayList())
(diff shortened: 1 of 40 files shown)


reaktive/src/commonTest/kotlin/com/badoo/reaktive/utils/SharedListIteratorTest.kt
```diff
@@ -0,0 +1,61 @@
...
+class SharedListIteratorTest {
+
+    @Test
+    fun listIterator() {
+        val lists: Pair<MutableList<Int?>, MutableList<Int?>> = Pair(SharedList(), ArrayList())
+        lists.first.addAll(listOf(0, null, 1, null, 2))
+        lists.second.addAll(listOf(0, null, 1, null, 2))
+        val iterators = Pair(lists.first.listIterator(), lists.second.listIterator())
+
+        iterators.verifyState()
```
CherryPerry: This test looks bizarre.

PR #215, opened 2019-10-01
Strict freezing in tests

arkivanov:


reaktive-testing/src/commonMain/kotlin/com/badoo/reaktive/test/maybe/TestMaybe.kt
```diff
@@ -4,9 +4,14 @@ import com.badoo.reaktive.maybe.Maybe
 import com.badoo.reaktive.maybe.MaybeCallbacks
 import com.badoo.reaktive.maybe.MaybeObserver
 import com.badoo.reaktive.test.base.TestSource
+import com.badoo.reaktive.utils.freeze
 
 class TestMaybe<T> : TestSource<MaybeObserver<T>>(), Maybe<T>, MaybeCallbacks<T> {
 
+    init {
+        freeze()
```
CherryPerry: Maybe move it into `TestSource` to avoid duplication?

arkivanov: It will work for now, but if you will add a field later into descendant class it will crash, because the class is already frozen. What do you think?

arkivanov: And it won't work for TestObserver as we already have fields in descendants.

CherryPerry: Then keep it as it is now.

arkivanov:
@CherryPerry Fixed the tests,

PR #216, opened 2019-10-02
Annotate _isPrintErrorEnabled with @SharedImmutable

arkivanov:
Same as in `Schedulers.kt`, otherwise it's impossible to access this field from non main threads.


build.gradle
```diff
@@ -1,5 +1,5 @@
 buildscript {
-    ext.reaktive_version = '1.0.0-rc2'
+    ext.reaktive_version = '1.0.0'
```
CherryPerry: Do in separate ticket!
Update Readme (operators, coroutines-interop).

Issue #213, opened 2019-10-01
Enable virtualization back for iOS


CherryPerry:
See [IosPlugin](https://github.com/badoo/Reaktive/blob/master/buildSrc/src/main/kotlin/IosPlugin.kt#L43).

PR #217, opened 2019-10-02
Enable devirtualization back for iOS

CherryPerry:
Fixes https://github.com/badoo/Reaktive/issues/213

If CI fails, do not close PR. Let's wait for next Kotlin version.

Ticket: [KT-34427](https://youtrack.jetbrains.com/issue/KT-34427)
Should be fixed in 1.3.60.


CherryPerry:
Fixed.

PR #220, opened 2019-10-06
Build samples on CI

CherryPerry:
@arkivanov Check Linux sample please.


arkivanov:
Linux sample is fine, actually I don't see any difference in error dialog.
image

Issue #224, opened 2019-10-07
Check linuxArm32Hfp target on CI


arkivanov:
Looks like `check` task does not involve `linuxArm32Hfp` target.

PR #225, opened 2019-10-08
Use build task instead of check

CherryPerry:
Fixes https://github.com/badoo/Reaktive/issues/224

Build = compile everything + check

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,12 +14,12 @@ matrix:
           - build-tools-28.0.3
           - android-28
       env: MP_TARGET=COMMON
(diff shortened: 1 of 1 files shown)


arkivanov:
But what about tests. Looks like they all are skipped now. And there is no test task involved for linuxArm32Hfp

CherryPerry:
@arkivanov `linuxX64Test` tasks were executed, but no output is available. Looks like test's results were cached, I didn't change code. In the last `master` [build](https://travis-ci.org/badoo/Reaktive/jobs/594836252) there is no test output too.

We can't execute tests for `arm` on `x86_64`. Only `LinuxX64` is tested.
The same for iOS builds: we test only `iosSim`, but not `iosArm32` and  `iosArm64`.

arkivanov:
Why? I can run linuxArm32 tests locally on my Ubuntu host. "Linux MIPS targets (linuxMips32 and linuxMipsel32) require a Linux host. Other Linux targets can be built on any supported host;"

CherryPerry:
Can you show me your Gradle tasks of `verification` group of `:reaktive` module?

arkivanov:
Hmm, looks like I'm wrong. There is no linux arm 32 test task there.

Issue #204, opened 2019-09-20
Detekt


CherryPerry:
Setup detekt for project?
Probably we also need to tweak to fix at least "No new line at the end of file".

arkivanov:
I agree

PR #227, opened 2019-10-08
Configure detekt

CherryPerry:
Fixes https://github.com/badoo/Reaktive/issues/204

Still long path to fix / ignore every issue.


CherryPerry:
Suppressed some issues.

CherryPerry:
Merge after https://github.com/badoo/Reaktive/pull/233

CherryPerry:
Ready to merge.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/CombineLatest.kt
```diff
@@ -127,6 +129,7 @@ fun <T1, T2, T3, T4, R> combineLatest(
             mapper(values[0] as T1, values[1] as T2, values[2] as T3, values[3] as T4)
         }
 
+@Suppress("LongParameterList")
```
arkivanov: Let's disable this inspection please

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/Zip.kt
```diff
@@ -9,6 +11,7 @@ import com.badoo.reaktive.utils.atomic.update
 import com.badoo.reaktive.utils.replace
 import com.badoo.reaktive.utils.serializer.serializer
 
+@Suppress("ComplexMethod")
```
arkivanov: And perhaps this should be disabled as well?

reaktive/src/jvmJsCommonMain/kotlin/com/badoo/reaktive/utils/serializer/SerializerImpl.kt
```diff
@@ -9,6 +9,7 @@ internal abstract class SerializerImpl<in T>(queue: Queue<T>) : Serializer<T> {
     private val monitor = Any()
     private var isDraining = false
 
+    @Suppress("ReturnCount")
```
arkivanov: What about this one? Do we really want it?

CherryPerry:
I wanted to discuss it in person, but forgot.
Main point: all these checks are based on some experience, and I think it is good to at least try to follow them. In places, where we can't for some reason follow them - suppress.

`LongParameterList` for example is suppressed only in operators public interface.
But `ComplexMethod` and `ReturnCount` could be fixed in future.

reaktive/src/jvmJsCommonMain/kotlin/com/badoo/reaktive/utils/serializer/SerializerImpl.kt
CherryPerry: `return X` count in one method.

Issue #230, opened 2019-10-10
Introduce time constants in Utils package


CherryPerry:
There are a lot of time constants we use across app (search for `1_000L` or `1_000_000L`).
We should extract them to single place.

PR #233, opened 2019-10-11
Extract time constants

EuV:
There were two approaches in naming convention:

- `SECOND_IN_NANOS` – "Convert seconds to nanoseconds" (I guess)
- `NANOS_IN_SECOND` – "How many nanoseconds in a second?"

I used the second one everywhere.

#230


utils/src/commonMain/kotlin/com/badoo/reaktive/utils/Constants.kt
```diff
@@ -0,0 +1,10 @@
+package com.badoo.reaktive.utils
+
+const val NANOS_IN_MICROS = 1_000L
+const val NANOS_IN_MILLIS = 1_000_000L
```
arkivanov: I believe the last word should be singular: `NANOS_IN_MICRO`, `NANOS_IN_MILLI`, `MICROS_IN_MILLI`. I have googled both variants and this one is being used more often.

PR #237, opened 2019-10-14
Split toList for native and non-native

arkivanov:
It's to expensive to allocate a new list for every item. We can allocate new lists only in Native and only if it's frozen.


reaktive/src/nativeCommonTest/kotlin/com/badoo/reaktive/observable/ToListTestNative.kt
```diff
@@ -0,0 +1,24 @@
...
+    @Test
+    fun collects_all_items_without_freezing() {
+        val upstream = TestObservable<Int?>(autoFreeze = false)
+        val observer = upstream.toList().test(autoFreeze = false)
+
+        upstream.onNext(0, null, 1, null, 2)
+        upstream.onComplete()
+
+        observer.assertSuccess(listOf(0, null, 1, null, 2))
+        assertFalse(observer.value.isFrozen)
```
CherryPerry: Do we need `isFrozen` as part of our util package and available for all platforms? Like `freeze()` or `ensureNeverFrozen()`? Just return `false` for all platforms except native.

In this case we will have single test implementation for all platforms.

reaktive/src/nativeCommonMain/kotlin/com/badoo/reaktive/observable/ToList.kt
```diff
@@ -0,0 +1,19 @@
+package com.badoo.reaktive.observable
+
+import com.badoo.reaktive.single.Single
+import kotlin.native.concurrent.isFrozen
+
+actual fun <T> Observable<T>.toList(): Single<List<T>> =
+    collect(ArrayList()) { list, item ->
+        val newList =
+            if (list.isFrozen) {
```
CherryPerry: If `list.isFrozen` is always false, then we will have following behaviour:
```kotlin
val newList = list
newList.add(item)
newList
```

Which is absolutely the same as `jvmJsCommon` implementation. Merge them?

reaktive/src/nativeCommonTest/kotlin/com/badoo/reaktive/observable/ToListTestNative.kt
```diff
@@ -0,0 +1,21 @@
...
+
+    @Test
+    fun collects_all_items_with_freezing() {
+        val upstream = TestObservable<Int?>()
+        val observer = upstream.toList().test()
+
+        upstream.onNext(0, null, 1, null, 2)
+        upstream.onComplete()
+
+        observer.assertSuccess(listOf(0, null, 1, null, 2))
```
CherryPerry: `assertTrue(observer.value.isFrozen)`?

Issue #203, opened 2019-09-20
Retrofit adapter


AlexeyKorshun:
Where do you think it would be better to put the adapter for retrofit, in this repository or in theirs?

CherryPerry:
I don't think they will accept this adapter.
And I am not sure about adding it here. We don't use Retrofit, so I think this adapter will become unmaintained in the nearest future.

If you need this adapter, you can host it on your account, and we will mention you in readme.

@arkivanov What do you think?

arkivanov:
I would try adding it to Retrofit repo. If rejected then add it here. The feature is nice to have anyway.

arkivanov:
Are there any news?

AlexeyKorshun:
@arkivanov at the moment there is no news. I try to create PR to square repository in next weekends.

AlexeyKorshun:
@arkivanov I asked them about it, because  their repository doesn't contain Kotlin and doesn't contain coroutines adapter. I think it may be a reason.

AlexeyKorshun:
Jake didn't appreciate the idea. [issue](https://github.com/square/retrofit/issues/3227)

AlexeyKorshun:
i have a trouble with using reaktive module in java module. I have warning 'java module can't use android dependency" and i can to create test. do you know how to fix this problem?

arkivanov:
Could you please show your build.gradle or pom.xml ?

arkivanov:
Try to specify the dependency as implementation project(':reaktive-jvm')

AlexeyKorshun:
> Try to specify the dependency as implementation project(':reaktive-jvm')

it don't help 
`Project with path ':reaktive-jvm' could not be found`
I understand this warning isn't important. But when I execute a test it says that can't resolve dependancy from reaktive module.

AlexeyKorshun:
I will try to resolve dependency from maven, but I m not sure that is great.

arkivanov:
I deleted my comment as it

> > Try to specify the dependency as implementation project(':reaktive-jvm')
> 
> it don't help
> `Project with path ':reaktive-jvm' could not be found`
> I understand this warning isn't important. But when I execute a test it says that can't resolve dependancy from reaktive module.

The warning itself is not important and it works for rxjava interop modules. Tests are also fine. Please double check that everything is same.

arkivanov:
Dependency should be specified as `project('reaktive')`, sorry for confusions

arkivanov:
[...]

PR #238, opened 2019-10-18
Added retrofit module

AlexeyKorshun:
#203 retrofit call adapter factory for reaktive library.


retrofit/src/main/kotlin/com/badoo/reaktive/retrofit/ReaktiveCallAdapter.kt
```diff
@@ -0,0 +1,36 @@
...
+
+import com.badoo.reaktive.observable.Observable
+import com.badoo.reaktive.observable.asCompletable
+import com.badoo.reaktive.observable.firstOrComplete
+import com.badoo.reaktive.observable.firstOrError
+import retrofit2.Call
+import retrofit2.CallAdapter
+import java.lang.reflect.Type
+
+internal class ReaktiveCallAdapter(
```
arkivanov: Can be replaced with just:
```
internal inline fun <T, R> callAdapter(responseType: Type, crossinline adapt: (Call<T>) -> R): Any =
    object : CallAdapter<T, R> {
        override fun responseType(): Type = responseType

        override fun adapt(call: Call<T>): R = adapt(call)
    }  
```

retrofit/src/main/kotlin/com/badoo/reaktive/retrofit/ReaktiveCallAdapter.kt
```diff
@@ -0,0 +1,36 @@
...
+        private val responseType: Type,
+        private val isBody: Boolean,
```
arkivanov: This set of flags smell, it is possible to specify all of them or any combination. Enum?

retrofit/src/main/kotlin/com/badoo/reaktive/retrofit/ReaktiveCallAdapterFactory.kt
```diff
@@ -0,0 +1,58 @@
...
+                check(observableType is ParameterizedType) { "Response must be parameterized as Response<Foo> or Response<? extends Foo>" }
+                responseType = getParameterUpperBound(0, observableType)
+            }
+            else -> {
+                responseType = observableType
+                isBody = true
+            }
+        }
+
+        return ReaktiveCallAdapter(responseType, isBody, isSingle, isMaybe, false)
```
arkivanov: All the code is so Java'ish, here is an idea
```
when {
    rawType == Completable::class.java -> callAdapter(responseType) { it.asSingle().asCompletable() }
    rawType == Single::class.java -> callAdapter(responseType) { it.asSingle() }
    rawType == Maybe::class.java -> callAdapter(responseType) { it.asSingle().asMaybe() }
    rawType == Observable::class.java -> callAdapter(responseType) { it.asSingle().asObservable() }
}
```

settings.gradle
```diff
@@ -15,4 +15,5 @@ if (target.isCommon()) {
     include ':sample-linuxx64-app'
     include ':rxjava2-interop'
     include ':rxjava3-interop'
+    include ':retrofit'
```
arkivanov: I'm not sure that this is a good module name, @CherryPerry what do you think?

CherryPerry:
So, my original opinion was to keep this adapter almost same as in retrofit repository, but I missed, that it is still written in java (even if OkHttp was already migrated to Kotlin).

So I think, this module should be written in more Kotlin-way, not just copy-paste, see @arkivanov proposals.

settings.gradle
CherryPerry: Yes, I think `retrofit-adapter` will be much better.

retrofit/src/test/kotlin/com/badoo/reaktive/retrofit/StringConverterFactory.kt
```diff
@@ -0,0 +1,28 @@
...
+import retrofit2.Retrofit
+import java.lang.reflect.Type
+
+internal class StringConverterFactory : Converter.Factory() {
+
+    override fun responseBodyConverter(
+            type: Type,
+            annotations: Array<Annotation>,
+            retrofit: Retrofit
+    ): Converter<ResponseBody, String> {
```
CherryPerry: Single return function.
Convert to 
```kotlin
): Converter<ResponseBody, String> =
    Converter { responseBody -> responseBody.string() }
```

retrofit/src/main/kotlin/com/badoo/reaktive/retrofit/CallExecuteObservable.kt
```diff
@@ -0,0 +1,48 @@
+package com.badoo.reaktive.retrofit
+
+import com.badoo.reaktive.disposable.Disposable
+import com.badoo.reaktive.observable.Observable
+import com.badoo.reaktive.observable.ObservableObserver
+import retrofit2.Call
+import retrofit2.Response
+
+internal class CallExecuteObservable<T>(private val originalCall: Call<T>) : Observable<Response<T>> {
```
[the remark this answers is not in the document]
CherryPerry: I don't remember implementation details, but is not `Call` is synchronous?
Looks like you need to implement [CallEnqueueObservable](https://github.com/square/retrofit/blob/master/retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/CallEnqueueObservable.java) and not [CallExecuteObservable](https://github.com/square/retrofit/blob/master/retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/CallExecuteObservable.java). You can add scheduler to factory with `ioScheduler` as default, but I think it is better to reuse thread pool of `OkHttpClient`.
Also you don't need to implement `Callback<Response>` as separate class, use solution provided above.

retrofit/src/main/kotlin/com/badoo/reaktive/retrofit/BodyObservable.kt
```diff
@@ -0,0 +1,42 @@
+package com.badoo.reaktive.retrofit
+
+import com.badoo.reaktive.disposable.Disposable
+import com.badoo.reaktive.observable.Observable
+import com.badoo.reaktive.observable.ObservableObserver
+import retrofit2.HttpException
+import retrofit2.Response
+
+internal class BodyObservable<T>(private val upstream: Observable<Response<T>>) : Observable<T?> {
```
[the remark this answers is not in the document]
CherryPerry: I think that `Single` will be better too. And then convert it to `Observable`, if someone for some strange reason want to use `Observable`.

retrofit/build.gradle
```diff
@@ -0,0 +1,25 @@
...
+        jvm(MavenPublication) {
+            from components.java
+        }
+    }
+}
+
+dependencies {
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
+    implementation project(':reaktive')
+    implementation project(':reaktive-testing')
```
CherryPerry: `testImplementation`? What is the reason to bundle testing utils with main package?

retrofit/src/test/kotlin/com/badoo/reaktive/retrofit/StringConverterFactory.kt
CherryPerry: Also, isn't it easier just to add dependency on `converter-scalars`?

AlexeyKorshun:
@arkivanov @CherryPerry sorry i had not enought time, i am fixing the bugs in this week.

CherryPerry:
Don't worry, take your time.

retrofit/src/main/kotlin/com/badoo/reaktive/retrofit/CallExecuteObservable.kt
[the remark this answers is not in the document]
AlexeyKorshun: @CherryPerry yes `Call` is synchronous. Why do we need to use Callback?

retrofit/src/test/kotlin/com/badoo/reaktive/retrofit/StringConverterFactory.kt
AlexeyKorshun: I didn't known about this library. Thx for information.

retrofit/src/main/kotlin/com/badoo/reaktive/retrofit/CallExecuteObservable.kt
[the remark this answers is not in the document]
CherryPerry: So, original adapter [factory](https://github.com/square/retrofit/blob/master/retrofit-adapters/rxjava2/src/main/java/retrofit2/adapter/rxjava2/RxJava2CallAdapterFactory.java) has 3 methods: `create()`, `createAsync()` and `createWithScheduler()`. I am fine with implementing only one of them, but not `create()`. It is absolutely inconvenient API that makes you set scheduler for each call! `createAsync()` uses OkHttp's tread pool via using Callback API. `createWithScheduler()` uses Call API, but also set valid scheduler via `subscribeOn()`.

So, the best option if you implement all of them.
Or you can just implement `create()` and `createWithScheduler()`.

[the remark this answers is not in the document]
AlexeyKorshun: Hmm. Ok. I know about all of them. But I think `create()` is more flexible way. I'll add `createWithSchedule()` method in this PR and will implement `createAsync()` in separate. @CherryPerry @arkivanov Will that be ok?

retrofit-adapter/src/main/kotlin/com/badoo/reaktive/retrofit/ReaktiveCallAdapterFactory.kt
```diff
@@ -0,0 +1,62 @@
...
+            Completable::class.java -> true
+            else -> {
+                check(this is ParameterizedType) { "Return type must be parameterized" }
+                getRawType(getParameterUpperBound(0, this)) != Response::class.java
+            }
+        }
+    }
+}
+
+private fun <T> Single<Response<T>>.transformToBody(transform: Boolean): Single<*> {
```
CherryPerry: It is single return function.
```
... Single<*> = 
    if (transform) ... then ...
```

Also can you put this function inside `ReaktiveCallAdapterFactory`?

CherryPerry:
Add `retrofit-adapter:build` step to Linux build inside `.github/workflows/build.yml` please.

retrofit/src/main/kotlin/com/badoo/reaktive/retrofit/CallExecuteObservable.kt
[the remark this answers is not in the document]
CherryPerry: Yes, it is fine.

AlexeyKorshun:
@CherryPerry @arkivanov good day. I fixed the comments, but CI reports a problem with failed tests. The tests were successful at the local machine. But when I cleared my cache I got the 'NoClassDefFoundError: com/badoo/reaktive/test/base/TestObserver' error again. Do you have any ideas what might be wrong?

colriot:
Don't you think that this should be a separate thing? Reaktive is about managing Rx streams, not about network or whatever. At the end of the day, you wouldn't want to keep and maintain everything related to Reaktive in the main repo, right? Imagine adapters for Room, Retrofit, SQLDelight, and lots of others living here.

arkivanov:
@CherryPerry good point above, what do you think? E.g. there is LiveData interop for Reaktive here in a separate repo https://github.com/Lamartio/Reaktive-LiveData-Interop

CherryPerry:
@arkivanov Wrote about it in original issue: https://github.com/badoo/Reaktive/issues/203#issuecomment-533624206. But you decided to add it here.

arkivanov:
Ok my fault, sorry about that. @AlexeyKorshun unfortunately we are not going to accept the adapter for the reasons above. Also we don't want this to become unmaintained as we are not using Retrofit. Feel free to publish it in your or any other repository. We would love to see it published and used by community.

PR #243, opened 2019-10-18
Split DisposableWrapper for Native and JVM/JS

arkivanov:


reaktive/src/commonMain/kotlin/com/badoo/reaktive/disposable/DisposableWrapper.kt
```diff
@@ -1,46 +1,21 @@
...
-import com.badoo.reaktive.utils.atomic.getAndUpdate
-
 /**
  * Thread-safe container of one [Disposable]
  */
-class DisposableWrapper : Disposable {
-
-    private val ref = AtomicReference<Holder?>(Holder(null))
-    override val isDisposed: Boolean get() = ref.value == null
+@Suppress("EmptyDefaultConstructor")
```
CherryPerry: Quite strange, that we need to suppress it.
https://github.com/arturbosch/detekt/blob/master/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/empty/EmptyDefaultConstructor.kt#L30

reaktive/src/jvmJsCommonMain/kotlin/com/badoo/reaktive/disposable/DisposableWrapper.kt
```diff
@@ -0,0 +1,51 @@
...
+    private var disposable: Disposable? = null
+
+    /**
+     * Disposes this [DisposableWrapper] and a stored [Disposable] if any.
+     * Any future [Disposable] will be immediately disposed.
+     */
+    actual override fun dispose() {
+        val disposableToDispose: Disposable?
+
+        synchronized(this) {
```
CherryPerry: Smaller implementation, but probably less readable.
```kotlin
synchronized(this) {
    _isDisposed = true
    disposable.also { disposable = null }
}?.dispose()
```

reaktive/src/jvmJsCommonMain/kotlin/com/badoo/reaktive/disposable/DisposableWrapper.kt
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * Atomically either replaces any existing [Disposable]
+     * with the specified one or disposes it if wrapper is already disposed.
+     * Also disposes any replaced [Disposable].
+     */
+    actual fun set(disposable: Disposable?) {
+        val disposableToDispose: Disposable?
+
+        synchronized(this) {
```
CherryPerry: Same can be done here.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/disposable/DisposableWrapper.kt
CherryPerry: Can you try `expect class DisposableWrapper expect construction()`?

arkivanov: `expect` is not allowed for constructors :-(

CherryPerry: Try just constructor then?

arkivanov: Still fails

PR #244, opened 2019-10-18
Split Disposable builders for Native, JVM and JS

arkivanov:


arkivanov:
I'm not 100% sure about this PR. Splitting for "with callback" and "without callback" variants allows to use method references like `::Disposable`. Also perhaps `Disposable()` is a better naming than `disposable()`, like `List()` or `MutableList()`.

PS: Second commit is just a refactoring.

CherryPerry:
> `Disposable { }`

Don't like it. I think, usage is the same as `observable { }`.
Also you don't instantiate class as expected, you use function. And you expect that function name starts with lower case. I saw same approach for `SharedList`, but decided to keep it as is, because it is almost internal stuff.

reaktive/src/commonTest/kotlin/com/badoo/reaktive/disposable/DisposableBuilderWithCallbackTest.kt
```diff
@@ -0,0 +1,46 @@
...
+        val disposable = Disposable {}
+
+        disposable.dispose()
+
+        assertTrue(disposable.isDisposed)
+    }
+
+    @Test
+    fun calls_callback_WHEN_disposed() {
+        var isCalled = false
```
CherryPerry: We use `AtomicBoolean()` in tests for such cases, check `DoOnBeforeErrorTest` as example.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/disposable/Various.kt
```diff
@@ -1,16 +1,14 @@
...
-inline fun disposable(crossinline onDispose: () -> Unit = {}): Disposable =
-    object : Disposable {
-        @Suppress("ObjectPropertyName") // Backing property
-        private var _isDisposed = AtomicBoolean()
-        override val isDisposed: Boolean get() = _isDisposed.value
+@Deprecated(
+    message = "Use Disposable instead",
+    replaceWith = ReplaceWith("Disposable(onDispose)", "com.badoo.reaktive.disposable.Disposable")
+)
+@JsName("disposableDeprecated")
```
CherryPerry: It is breaking change. You will break current JS code.

arkivanov:
> > `Disposable { }`
> 
> Don't like it. I think, usage is the same as `observable { }`.
> Also you don't instantiate class as expected, you use function. And you expect that function name starts with lower case. I saw same approach for `SharedList`, but decided to keep it as is, because it is almost internal stuff.

It is like `List(size) { }` or `Array(size) {}`. Builder function names are allowed to start from capital letter.

CherryPerry:
I don't like this change. Looks very overcomplicated for such simple staff.

If you want to use method reference then maybe just remove default value from `disposable` function and introduce function as you did, without splitting?

```kotlin
inline fun disposable(crossinline onDispose: (() -> Unit)?) { ... }
fun disposable() = disposable { }
```

Also you can make `onDispose` nullable.

You don't need to split implementations of disposable with callback, jvm and native implementations are almost the same! I think performance increase in js implementation is too low to maintain it separately.

arkivanov:
The main thing is that JVM and js variants now produce only one allocation instead of two (no AtomicBoolean is used). You can make lambda billable for inline functions. Ability to use method references is a minor side effect. Also inlining anonymous class generates a class per call.

arkivanov:
This change increases Scrabble benchmark results by 10%

arkivanov:
Actually at the moment there are 3 allocations per disposable (anonymous class itself + our AtomicBooolean + actual AtomicBoolean)

reaktive/src/commonMain/kotlin/com/badoo/reaktive/disposable/Various.kt
arkivanov: I'm still not 100% sure what exactly it breaks. I have checked our sample-js-app - the deprecated method is still visible and available as `disposable` and marked as deprecated.

reaktive/src/commonTest/kotlin/com/badoo/reaktive/disposable/DisposableBuilderWithCallbackTest.kt
arkivanov: Because of freezing by our `TestObserver`, `TestSource`, etc.. There is no freezing here at the moment. Why do you want to use `AtomicBoolean`?

reaktive/src/commonMain/kotlin/com/badoo/reaktive/disposable/Various.kt
CherryPerry: If someone uses `disposable` in js code, it won't work anymore. Even if no one uses it, in future it is possible, should be careful with such changes.

arkivanov: Should I revert the name change? AFAIK Kotlin functions with arguments are exported with mangled names like 'repeat_j15sdp'.

CherryPerry: Keep it, you can mention name change in changelog.

PR #245, opened 2019-10-20
Use buildSrc for dependency management

CherryPerry:
So, I created `buildSrc` folder for `buildSrc` and it works.

image


arkivanov:
image

buildSrc/buildSrc/src/main/kotlin/Deps.kt
```diff
@@ -0,0 +1,67 @@
...
+        object Test {
+            const val common = "org.jetbrains.kotlin:kotlin-test-common:$kotlinVersion"
+            const val js = "org.jetbrains.kotlin:kotlin-test-js:$kotlinVersion"
+            const val junit = "org.jetbrains.kotlin:kotlin-test-junit:$kotlinVersion"
+            const val annotationsCommon = "org.jetbrains.kotlin:kotlin-test-annotations-common:$kotlinVersion"
+        }
+    }
+
+    object Kotlinx {
+        val coroutines = Coroutines
```
arkivanov: Why do we need such fields? Why not to reference object directly? Like `Deps.Kotlinx.Coroutines.blah`. Currently you can use both ways btw.

buildSrc/buildSrc/src/main/kotlin/Deps.kt
```diff
@@ -0,0 +1,67 @@
...
+        val plugin = "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion"
+
+        class Stdlib(
+            private val name: String = "org.jetbrains.kotlin:kotlin-stdlib:$kotlinVersion"
+        ) : CharSequence by name {
+            val jdk7 = "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlinVersion"
+            val js = "org.jetbrains.kotlin:kotlin-stdlib-js:$kotlinVersion"
+            val common = "org.jetbrains.kotlin:kotlin-stdlib-common:$kotlinVersion"
+
+            override fun toString(): String = name
```
arkivanov: Why here we have `name` and `toString` and `CharSequence` and don't have it for e.g. `Test`?

buildSrc/buildSrc/src/main/kotlin/Deps.kt
arkivanov: Can we make them objects everywhere so they will be like groups?

buildSrc/buildSrc/src/main/kotlin/Deps.kt
CherryPerry: You can use both `Deps.kotlin.sdtlib` and `Deps.kotlin.stlib.common`.
`implementation` call accepts `CharSequence`, so to achieve behaviour above, I implemented `CharSequence` interface on `Stdlib` class and provided `name` property for interface delegation.
Otherwise you need to introduce different field on `Stdlib` object for `org.jetbrains.kotlin:kotlin-stdlib` and API won't be so cool (like `Deps.kotlin.sdtlib.default`).

buildSrc/buildSrc/src/main/kotlin/Deps.kt
CherryPerry: > Why do we need such fields?

You can't use `Deps.Kotlinx.Coroutines.blah` in Groovy. Only via `Deps$Kotlinx`.

CherryPerry: > Can we make them objects everywhere so they will be like groups?

Did not get it. Sample?

buildSrc/buildSrc/src/main/kotlin/Deps.kt
CherryPerry: `toString()` method is not delegated to `name`, so I have to implement it manually. Gradle calls `toString()` on this object somewhere inside.

PR #250, opened 2019-10-21
Introduce ObjectReference for simple cases

arkivanov:
There are circumstances when `AtomicReference` is not actually required and is used only because of K/N freezing things. In this PR I introduce _internal_ `ObjectReference` which is a simple mutable holder in JVM and JS and is still backed by `FreezableAtomicReference` in K/N.


reaktive/src/commonMain/kotlin/com/badoo/reaktive/utils/ObjectReference.kt
```diff
@@ -0,0 +1,6 @@
+package com.badoo.reaktive.utils
+
+internal expect class ObjectReference<T>(initialValue: T) {
```
CherryPerry: Maybe make it public? Can be useful to write own operators. Just add some kdoc and alert like "if you are not sure, use AtomicReference".

PR #261, opened 2019-10-23
Optimized Observable flatMap

arkivanov:


reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/FlatMap.kt
```diff
@@ -26,7 +24,7 @@ fun <T, R> Observable<T>.flatMap(mapper: (T) -> Observable<R>): Observable<R> =
                     }
 
                 override fun onSubscribe(disposable: Disposable) {
-                    disposables += disposable
+                    add(disposable)
```
CherryPerry: Do same base class as for `DisposableWrapper`? Can be useful in future.

PR #266, opened 2019-10-27
Build with GitHub Actions

CherryPerry:
Generic checks on Actions, release on Travis-CI.

Just approve, don't merge. Want to double check something else.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,6 @@
 name: Build
 
-on: [push]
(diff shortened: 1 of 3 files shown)


CherryPerry:
Quite sad, that CI changes in this PR are not applied to check phase. But changes in `.travis.yml` are applied instantly.

Maybe they will in future, after first Action is applied.

Build in my repo: https://github.com/CherryPerry/Reaktive/runs/276528376

PR #268, opened 2019-10-30
Copy-on-add Map in `toMap` operator only on Native and if frozen

colriot:
Skip dances with freezing when it's not valid for the platform or use-case


reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/ToMap.kt
```diff
@@ -1,13 +1,22 @@
...
 fun <T, K> Observable<T>.toMap(keySelector: (T) -> K): Single<Map<K, T>> =
     toMap(keySelector) { it }
 
 fun <T, K, V> Observable<T>.toMap(keySelector: (T) -> K, valueSelector: (T) -> V): Single<Map<K, V>> =
-    collect(emptyMap()) { map, item ->
+    collect(LinkedHashMap()) { map, item ->
         val key = keySelector(item)
         val value = valueSelector(item)
-        map.plus(key to value)
+        if (map.isFrozen) {
```
colriot: I'm wondering whether we can safely assume that if `map` is not frozen it will stay like that during collection. If "yes" then we can move `if` check outside of the loop and do it only once.

arkivanov: It can be frozen by an upstream. You can try, just remove `if` and run the `collects_all_items_with_freezing` test for Linux or iOS. This check is necessary.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/ToMap.kt
```diff
@@ -1,13 +1,22 @@
...
 
 import com.badoo.reaktive.single.Single
+import com.badoo.reaktive.utils.isFrozen
 
 fun <T, K> Observable<T>.toMap(keySelector: (T) -> K): Single<Map<K, T>> =
     toMap(keySelector) { it }
 
 fun <T, K, V> Observable<T>.toMap(keySelector: (T) -> K, valueSelector: (T) -> V): Single<Map<K, V>> =
-    collect(emptyMap()) { map, item ->
+    collect(LinkedHashMap()) { map, item ->
```
arkivanov: Why `LinkedHashMap` and not just `HashMap`? RxJava uses `HashMap`.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/ToMap.kt
arkivanov: It is required. Upstream may freeze the list. You can try, just remove the 'if' and run native tests. One test will fail)

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/ToMap.kt
colriot: Because that's what Kotlin uses for `mapOf` and for copying `Map.plus`. If there are any benchmarks around occupied memory, we can check the real impact for optimization purposes. But you may want to check this out first: https://publicobject.com/2016/02/08/linkedhashmap-is-always-better-than-hashmap/

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/ToMap.kt
colriot: You mean, freeze in the middle of `collect`? I'll try to run native tests on my machine, last time I didn't succeed.

arkivanov: I mean upstream may freeze things in the middle of stream. There is no test for this particular case though. But you can write one for this operator and for 'toList' as well.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/ToMap.kt
arkivanov: Since the result is `Map` which is an interface there is no guarantee for iteration order. Actual map is implementation details. No one should assume anything about it. But lets use `LinkedHashMap` to be consistent in all cases.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/ToMap.kt
arkivanov: Could you please clarify what you mean by "move if check outside of the loop and do it only once"? Some code would help.

If I understood you correctly the following test should cover the case (it is for `toList`, you can adopt it for `toMap`):
```
    @Test
    fun collects_all_items_with_freezing_by_upstream_in_the_middle() {
        val upstream = TestObservable<Int>(autoFreeze = false)
        val observer = upstream.toList().test(autoFreeze = false)

        upstream.onNext(0)
        upstream.freeze()
        upstream.onNext(1)
        upstream.onComplete()

        observer.assertSuccess(listOf(0, 1))
        assertTrue(observer.value.isFrozen)
    }
```

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/ToMap.kt
colriot: I'm not assuming anything, I use the default Kotlin implementation. And article was shared FYI to get an idea why LHM might be used in Kotlin

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/ToMap.kt
colriot: Yes, you got me right. This test-case will fail my example. And I guess it's a valid use-case, like, when some items are passed through the stream without freezing, and some (by a predicate, for example) are redirected to another thread to be processed, and thus, these items end up being frozen.

colriot:
Added tests for the case of freezing in the middle of emission.

PR #272, opened 2019-10-31
Make freeze-related calls `inline`

colriot:
This will skip unnecessary method calls for essentially-aliased or constant methods. Also will allow JVM/JS compiler to eliminate unreachable code branches


utils/src/nativeCommonMain/kotlin/com/badoo/reaktive/utils/Freeze.kt
```diff
@@ -4,12 +4,13 @@ import kotlin.native.concurrent.ensureNeverFrozen
 import kotlin.native.concurrent.freeze
 import kotlin.native.concurrent.isFrozen
 
-actual fun <T> T.freeze(): T = freeze()
+@Suppress("NOTHING_TO_INLINE")
+actual inline fun <T> T.freeze(): T = freeze()
 
 actual fun <T: Any> T.ensureNeverFrozen(): T {
```
colriot: Probably this one is worth becoming `inline`d too

arkivanov: Yeah, let's do it. Also inline `isFrozen` please.

arkivanov: And better to annotate `expect` functions as inline as well

colriot: `isFrozen` is already inlined. Agree that need to move modifier to `expect` functions, thought about after pushing the PR too

colriot:
Updated with suggested changes

PR #273, opened 2019-10-31
Fix nitpicks

colriot:


reaktive/src/commonMain/kotlin/com/badoo/reaktive/maybe/DoOnBefore.kt
```diff
@@ -82,7 +82,7 @@ fun <T> Maybe<T>.doOnBeforeComplete(action: () -> Unit): Maybe<T> =
                 }
 
                 override fun onComplete() {
-                    tryCatch(action) {
+                    emitter.tryCatch(action) {
```
arkivanov: Let's do the same for `Observable` and `Completable`

colriot:
Updated with suggested changes

PR #278, opened 2019-11-02
Add more descriptive assertion messages

colriot:
* Cleanup assertion messages for testing API
* Move `isError` property inside `TestObserver`, as it's done with `isComplete`, `isSuccess`, etc.

### DIFF
--- a/reaktive-testing/src/commonMain/kotlin/com/badoo/reaktive/test/base/TestObserverExt.kt
+++ b/reaktive-testing/src/commonMain/kotlin/com/badoo/reaktive/test/base/TestObserverExt.kt
@@ -6,38 +6,36 @@ import kotlin.test.assertNotNull
 import kotlin.test.assertNull
 import kotlin.test.assertTrue
 
-val TestObserver.isError: Boolean get() = error != null
-
 fun TestObserver.assertError() {
-    assertTrue(isError)
+    assertTrue(isError, "Source did not fail")
 }
 
 fun TestObserver.assertError(expectedError: Throwable) {
-    assertEquals(expectedError, error)
+    assertEquals(expectedError, error, "Source error does not match")
 }
 
 fun TestObserver.assertError(predicate: (Throwable) -> Boolean) {
-    val error = this.error
-    assertNotNull(error)
-    assertTrue(predicate(error))
+    assertError()
+    val error = this.error!!
+    assertTrue(predicate(error), "Source error does not match the predicate: $error")
 }
 
 fun TestObserver.assertNotError() {
-    assertFalse(isError)
+    assertFalse(isError, "Source failed")
 }
 
 fun TestObserver.assertSubscribed() {
-    assertNotNull(disposable)
+    assertNotNull(disposable, "Source is not subscribed")
 }
 
 fun TestObserver.assertNotSubscribed() {
-    assertNull(disposable)
+    assertNull(disposable, "Source is subscribed")
 }
 
 fun TestObserver.assertDisposed() {
-    assertTrue(isDisposed)
+    assertTrue(isDisposed, "Source is not disposed")
 }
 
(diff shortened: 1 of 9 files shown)


reaktive-testing/src/commonMain/kotlin/com/badoo/reaktive/test/base/TestObserverExt.kt
```diff
@@ -6,38 +6,36 @@ import kotlin.test.assertNotNull
 import kotlin.test.assertNull
 import kotlin.test.assertTrue
 
-val TestObserver.isError: Boolean get() = error != null
```
arkivanov: This is kinda a breaking change. Also unlike `isComplete` and `isSuccess` this function can be extension as it uses only public API. `isComplete` and `isSuccess` both use private API.

colriot: Why did you make it an extension function at the first place? Maybe any special cases?

I actually don't see a good reason for that, all `isSmth` methods are analogous and, imho, should be parts of the internal API of the Observer.

arkivanov: Because I want to keep the main API as minimalistic as possible. If something can be an extension I would prefer this way. E.g. Observable failed so it has "error". This is part of the API. Then you can write extensions like isError, isFailed, isNotFailed as many as you want.

colriot: Well, imho, this example is actually far-fetched. `isError`, `isFailed` and `isNotFailed` are actually the same method. 

I completely agree with keeping the API minimalistic, but it doesn't mean moving things  to extension functions just because it is possible. I'd argue that `isError` has the same level of affiliation to the public API as `isComplete` and `isSuccess`. Because it's a straight mirroring of `*Observer`'s `onError`, `onSuccess`, `onComplete` methods. And it's strange to have supplements for latter two, but to not have one for the former one.

arkivanov: The main idea is: when a stream completes - the result is "completed" state. That's why we have is completed flag as part of the API. When a stream succeeds - the result is a value. Again it's part of the API. Same "mirroring" for error - when a stream fails, the state is "have Throwable". Again this is exposed via API as `error` property. So we have minimalistic API to check the state: isComplete, value, values and error. All others are extensions. Why do we have isSuccess as part of the API? Because nulls are allowed. So you can't check whether Single is succeeded or not based on value only. You need access to internals.

colriot: I understand this argumentation by **necessity** . I argue only with keeping the _only_ member of the API - `isError` - as an extension function. Because it, imho, also belongs to others as per my previous comment and analogy with `*Observer`s method. Here I'm following the _analogy_ principle – these methods are all about the same set or responsibilities (API for accessing Observer's state) and it's strange to divide them in the API.

arkivanov: Well we have a rule. If method is only using public API then it should be an extension. Unless there is any performance impact or something. The fact that this is the only such a method is just a coincidence. Also a public method that just only calls another own public method is sometimes considered as a code smell and anti-pattern.
Any way we are unable to accept the change as it is brakes binary compatibility and it's not worth it in this particular case.

arkivanov: Alright, we decided to accept the change. There is a chance we might need access to internals for this in future.

PR #279, opened 2019-11-02
Use library's own testing API

colriot:
* Use own testing API explicitly instead of basic `kotlin-test` asserts
* Move `assertSuccess` from implicit invocation in `value` getter to explicit usage in actual test methods. Making logic/intention explicit


reaktive-testing/src/commonMain/kotlin/com/badoo/reaktive/test/maybe/TestMaybeObserver.kt
```diff
@@ -9,13 +9,7 @@ import com.badoo.reaktive.utils.freeze
 class TestMaybeObserver<T>(autoFreeze: Boolean = true) : TestObserver(), MaybeObserver<T> {
 
     private val _value = AtomicReference<Value<T>?>(null)
-
-    val value: T
-        get() {
-            assertSuccess()
```
arkivanov: Why did you remove the assert? Now if `Maybe` not succeeded then NPE will be thrown which is much less understandable. This is the public API.

arkivanov: I think it's better to keep it, in case if a client will just read the value without `assertSuccess`.

colriot: Because that brings an implicit assumption that if you access `value` then some `assert` will be done for you (API user). And with tests it's always better to be explicit so that the test method itself will contain all the meaningful logic in its body. I.e. if a user assumes successful completion, they should assert it with `assertSuccess` explicitly.

I agree with non-clear NPE argument though. I'd use `requireNonNull(..., "Should 'assertSuccess' before accessing a 'value'")` which will explain and encourage explicit use of API. 
Wdyt?

arkivanov: I'm fine with requireNonNull, but I would like message to be more clear, like "Single did not success" or something. So it will be clear why there is no value. @CherryPerry what do you think?

colriot: How about "Single did not succeed. Assert that with 'assertSuccess()' before accessing a 'value'"?

arkivanov: Works for me. @CherryPerry ?

CherryPerry: Agree.

arkivanov: I have applied the requested changes

arkivanov:
@CherryPerry Had to fix the error text for `TestMaybeObserver`

PR #282, opened 2019-11-03
Inlined XxxByEmitter builders

arkivanov:


reaktive/src/commonMain/kotlin/com/badoo/reaktive/base/DisposableEmitter.kt
```diff
@@ -3,6 +3,7 @@ package com.badoo.reaktive.base
 import com.badoo.reaktive.disposable.Disposable
 import com.badoo.reaktive.disposable.DisposableWrapper
 
+@PublishedApi
```
arkivanov: Not sure about this. This annotation makes the class effectively public and it will be subject for binary compatibility. Maybe just remove?

arkivanov:
Thinking of a better way

arkivanov:
Will implement in another PR in a different way

PR #286, opened 2019-11-05
Introduce DisposableContainer

arkivanov:
- Deprecated `CompositeDisposable` (to avoid breaking changes)
- Implemented `DisposableContainer` with similar API as in RxJava

Commit 1: deprecated CompositeDisposable, introduced DisposableContainer, tests
Commit 2: replaced all usages

### DIFF
--- a/reaktive/src/jvmJsCommonMain/kotlin/com/badoo/reaktive/disposable/CompositeDisposable.kt
+++ b/reaktive/src/jvmJsCommonMain/kotlin/com/badoo/reaktive/disposable/CompositeDisposable.kt
@@ -8,13 +8,13 @@ import kotlin.jvm.Volatile
 @Suppress("EmptyDefaultConstructor")
 actual open class CompositeDisposable actual constructor() : Disposable {
 
-    private var list: MutableList<Disposable>? = null
+    private var collection: MutableCollection<Disposable>? = null
     @Volatile
     private var _isDisposed = false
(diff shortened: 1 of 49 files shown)


arkivanov:
Squashed to one commit and force pushed

reaktive/src/jvmJsCommonMain/kotlin/com/badoo/reaktive/disposable/CompositeDisposable.kt
```diff
@@ -8,13 +8,13 @@ import kotlin.jvm.Volatile
...
     @Volatile
     private var _isDisposed = false
     override val isDisposed: Boolean get() = _isDisposed
 
     /**
-     * Disposes the [CompositeDisposable] and all its [Disposable]s.
+     * Atomically disposes the collection and all its [Disposable]s.
      * All future [Disposable]s will be immediately disposed.
      */
     actual override fun dispose() {
```
CherryPerry: Just a question. What if we add kdoc to except class? Won't it be available on actual classes?
For example, if you write kdoc for interface methods, that it will be available on implementation methods.

arkivanov: Kdocs from expect classes are not visible, looks like a bug. So we have to copy them to actual classes(

Issue #241, opened 2019-10-18
Add IDEA files to repo


CherryPerry:
Check [video](https://www.droidcon.com/media-detail?video=362619508) before.

Main points:
- Continuation indent: 4 as all others
- No wildcard imports
- Line length same as in detekt

PR #290, opened 2019-11-09
Code style config for IDEA

CherryPerry:
Fixes https://github.com/badoo/Reaktive/issues/241

Please, open project in your environment before approve.


arkivanov:
Checked out your branch, here is the result:
image

PR #302, opened 2019-11-16
Introduce DisposableScope

arkivanov:
This is just an idea, I would like to hear your opinions, thoughts, etc. The idea is to ease disposable management. Also scopes can be nested.


reaktive/src/commonMain/kotlin/com/badoo/reaktive/disposable/DisposableScope.kt
```diff
@@ -0,0 +1,43 @@
...
+import com.badoo.reaktive.observable.Observable
+import com.badoo.reaktive.single.Single
+
+interface DisposableScope : Disposable {
+
+    fun <T : Disposable> T.scope(): T
+
+    fun cancel()
+
+    fun <T> Observable<T>.subscribeScoped(
```
CherryPerry: Can we do `Disposable.scoped()` instead of this? So I can do:
```kotlin
source.subscribe(...).scoped()
```

But it is easy to do mistake and use `scope()`, so we should rename it.

CherryPerry: I don't think we need them. What is the problem with `scope()`?

arkivanov:
Still WIP

reaktive/src/commonMain/kotlin/com/badoo/reaktive/disposable/DisposableScope.kt
arkivanov: Why do you think so? I find them very handy.

arkivanov:
@mk-jones @CherryPerry @zsoltk Could you please check the proposal? Please let me know you thoughts or any ideas.

arkivanov:
You can find some use cases in the updated samples. Another use case could be:
```
class MyView {
    fun showProgressBar() {
    }

    fun hideProgressBar() {
    }
}

class MyPresenter(
    private val view: MyView
) : DisposableScope by DisposableScope() {

    fun load() {
        view.showProgressBar()

        completableFromFunction {
            // A long running operation
        }
            .subscribeOn(ioScheduler)
            .observeOn(mainScheduler)
            .subscribeScoped(isThreadLocal = true, onComplete = view::hideProgressBar)
    }
}

class MyActivity : DisposableScope by DisposableScope() {

    fun onCreate() {
        +MyPresenter(MyView())
    }

    fun onDestroy() {
        dispose()
    }
}
```

reaktive/src/commonMain/kotlin/com/badoo/reaktive/disposable/DisposableScope.kt
CherryPerry: The problems with `subscribeScoped` are: they override current approach of using `subscribe` – it is much easier to try use them when you don't need or can't, and forget to use them when needed. `scope()` is more explicit in this way, especially in combination with `@CheckResult`. Also you need to support both subscribe variants.

sample-mpp-module/src/commonMain/kotlin/com/badoo/reaktive/samplemppmodule/store/KittenStoreImpl.kt
```diff
@@ -1,66 +1,54 @@
...
 import com.badoo.reaktive.subject.behavior.behaviorSubject
 import com.badoo.reaktive.utils.ensureNeverFrozen
 
 internal class KittenStoreImpl(
     private val loader: KittenLoader
-) : KittenStore {
+) : KittenStore, DisposableScope by DisposableScope() {
 
-    private val _states = behaviorSubject(State()).ensureNeverFrozen()
+    private val _states = +behaviorSubject(State()).ensureNeverFrozen()
```
CherryPerry: `+behaviorSubject(State())` feels like Compose...
Can't say yet, is it good or not.

reaktive/src/commonTest/kotlin/com/badoo/reaktive/disposable/scope/DisposableScopeImplTest.kt
```diff
@@ -0,0 +1,120 @@
...
+
+        scope.run { disposable.scope() }
+        scope.dispose()
+
+        assertTrue(disposable.isDisposed)
+    }
+
+    @Test
+    fun completes_scoped_CompleteCallback_WHEN_disposed() {
+        val completeCallback =
```
CherryPerry: Looks terrible. What is use case?

If you want to have some `onDispose` callback, then maybe it is better to do something like this?
```kotlin
scope.onDispose {
    ...
}
```

tools/binary-compatibility/reference-public-api/reaktive.txt
```diff
@@ -207,6 +207,33 @@ public final class com/badoo/reaktive/disposable/VariousJvm {
 	public static final fun Disposable (Lkotlin/jvm/functions/Function0;)Lcom/badoo/reaktive/disposable/Disposable;
 }
 
+public abstract interface class com/badoo/reaktive/disposable/scope/DisposableScope : com/badoo/reaktive/disposable/Disposable {
```
CherryPerry: Maybe extract to separate module? It is not a core functionality.

sample-mpp-module/src/commonMain/kotlin/com/badoo/reaktive/samplemppmodule/store/KittenStoreImpl.kt
ShikaSD: No good, they got rid of it :)

reaktive/src/commonMain/kotlin/com/badoo/reaktive/disposable/DisposableScope.kt
arkivanov: Well, If you are inside a scope then I hardly imagine that you will need just `subscribe` without cancellation (very rare case).
And if you forget to use `subscribeScoped` then you could forget to use `scope()` as well - but in both cases `@CheckResult` would help.

arkivanov: I just very like one with automatic disposable management

reaktive/src/commonMain/kotlin/com/badoo/reaktive/disposable/scope/DisposableScopeImpl.kt
```diff
@@ -0,0 +1,93 @@
...
+
+    override fun <T : CompleteCallback> T.scope(): T {
+        Disposable(::onComplete).scope()
+
+        return this
+    }
+
+    override fun <T : CompleteCallback> T.unaryPlus(): T = scope()
+
+    override fun <T> Observable<T>.subscribeScoped(
```
ShikaSD: Does Kotlin name resolution allow to name it `subscribe`?

reaktive/src/commonTest/kotlin/com/badoo/reaktive/disposable/scope/DisposableScopeImplTest.kt
arkivanov: This is something I'm not sure about. The use case is demonstrated in the sample, where `BehaviorSubject` is created and scoped using `unaryPlus` operator. So when the scope is cancelled - the subject is completed.

tools/binary-compatibility/reference-public-api/reaktive.txt
arkivanov: Perhaps, but I find it as a replacement for most of the `CompositeDisposable` use cases. So you will need it quite frequently. We could move `DisposableWrapper`, `CompositeDisposable` and `DisposableScope` later to a separate module. What do you think?

reaktive/src/commonMain/kotlin/com/badoo/reaktive/disposable/scope/DisposableScopeImpl.kt
arkivanov: It does, but I thought it will be very confusing to users.

sample-mpp-module/src/commonMain/kotlin/com/badoo/reaktive/samplemppmodule/store/KittenStoreImpl.kt
arkivanov: I can remove. But it is taken from the official docs and looks nice to me) https://kotlinlang.org/docs/reference/type-safe-builders.html

reaktive/src/commonTest/kotlin/com/badoo/reaktive/disposable/scope/DisposableScopeImplTest.kt
CherryPerry: Then maybe use `Subject` as base?

tools/binary-compatibility/reference-public-api/reaktive.txt
CherryPerry: If you want to use it inside operators, then it is fine to have it in main module.

reaktive/src/commonTest/kotlin/com/badoo/reaktive/disposable/scope/DisposableScopeImplTest.kt
arkivanov: You don't like the ability to complete a completable (something implementing CompletableCallback) whenever scope closes?

reaktive/src/commonMain/kotlin/com/badoo/reaktive/disposable/scope/DisposableScopeImpl.kt
ShikaSD: Not sure, but probably you're right

sample-mpp-module/src/commonMain/kotlin/com/badoo/reaktive/samplemppmodule/binder/KittenBinder.kt
```diff
@@ -20,21 +19,24 @@ class KittenBinder(
...
-                .events
-                .map(KittenViewEventToIntentMapper::invoke)
-                .subscribe(onNext = store::accept)
-
-        disposables +=
-            store
-                .states
-                .map(KittenStateToViewModelMapper::invoke)
-                .subscribe(onNext = { requireNotNull(view).show(it) })
+        startStopScope = disposableScope { start() }
```
mk-jones: It would be nice if the example used a non-nullable val for the scope, and then added to it like in the previous impl.

sample-mpp-module/src/commonMain/kotlin/com/badoo/reaktive/samplemppmodule/store/KittenStoreImpl.kt
mk-jones: Personally I would prefer a function call instead of the plus operator.
Plus operator isn't named, so the reader has to be actively aware of the context where it is called (it does different things in different places). I think plus operator works better in places where it is clear what you are adding to. Having to call scope at least tells you what is happening.
(particularly as people are likely reading the examples on github, and can't look into the definition of what + is adding to).

tools/binary-compatibility/reference-public-api/reaktive.txt
mk-jones: I think it's fine in the main module.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/disposable/scope/DisposableScopeBuilder.kt
```diff
@@ -0,0 +1,16 @@
...
+ * Creates a new instance of [DisposableScope]
+ */
+@Suppress("FunctionName")
+@JsName("disposableScope")
+fun DisposableScope(): DisposableScope = DisposableScopeImpl()
+
+/**
+ * Creates a new instance os [DisposableScope] and calls the provided `block` on it
+ */
+inline fun disposableScope(block: DisposableScope.() -> Unit): DisposableScope =
```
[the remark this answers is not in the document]
arkivanov: For what reason?

sample-mpp-module/src/commonMain/kotlin/com/badoo/reaktive/samplemppmodule/store/KittenStoreImpl.kt
arkivanov: Ok, looks like nobody like it. Will remove and use `.scope()` instead.

sample-mpp-module/src/commonMain/kotlin/com/badoo/reaktive/samplemppmodule/binder/KittenBinder.kt
arkivanov: How do you imagine this? `lateinit`? `onStart`-`onStop` can be called multiple times before `onDestroy`.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/disposable/DisposableScope.kt
mk-jones: Yep, looks convenient in my opinion.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/disposable/scope/DisposableScopeBuilder.kt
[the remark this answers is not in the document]
mk-jones: Don't remember, think I had misread this.

sample-mpp-module/src/commonMain/kotlin/com/badoo/reaktive/samplemppmodule/binder/KittenBinder.kt
mk-jones: I was thinking the same way with CompositeDisposable, with disposables.clear() in onStop.

arkivanov: Because it's lifecycle is between start and stop, I think better to keep it nullable and assign null on stop.

sample-mpp-module/src/commonMain/kotlin/com/badoo/reaktive/samplemppmodule/store/KittenStoreImpl.kt
arkivanov: Removed

reaktive/src/commonMain/kotlin/com/badoo/reaktive/disposable/scope/DisposableScopeBuilder.kt
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * Creates a new instance of [DisposableScope]
+ */
+@Suppress("FunctionName")
+@JsName("disposableScope")
+fun DisposableScope(): DisposableScope = DisposableScopeImpl()
+
+/**
+ * Creates a new instance os [DisposableScope] and calls the provided `block` on it
```
CherryPerry: > Creates a new instance **os** [DisposableScope]

Issue #285, opened 2019-11-03
Implement binary compatibility checks


CherryPerry:
https://github.com/Kotlin/kotlinx.coroutines/tree/master/binary-compatibility-validator
https://github.com/JetBrains/kotlin/tree/master/libraries/tools/binary-compatibility-validator

Pull request action: build, compare binary compatibility and post result as comment to pull request.
Master branch commit action: build, compare binary compatibility, commit if changed.

CherryPerry:
@arkivanov It is JVM only compatibility, is it fine?

PR #306, opened 2019-11-22
Binary compatibility checks

CherryPerry:
Copy-pasted solution from Kotlin repo. Don't change it, so it will be easy to sync it in future.
Checks only JVM compatibility.

Adopted files: `build.gradle.kts` and `RuntimePublicAPITest.kt`.

Fixes https://github.com/badoo/Reaktive/issues/285


.github/workflows/build.yml
```diff
@@ -31,3 +31,15 @@ jobs:
...
+    name: Binary compatibility
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v1
+      - name: Install Java
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+      - name: Run binrary compatibility test suite
```
arkivanov: Typo "binrary"

PR #307, opened 2019-11-23
Added Maybe.defaultIfEmpty alias

arkivanov:


CherryPerry:
@arkivanov Merge after my ticket please and run compatibility override.

arkivanov:
@CherryPerry binary compatibility check is failed. What should we do?

PR #308, opened 2019-11-24
Implement repeat operators

arkivanov:


reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/Repeat.kt
```diff
@@ -0,0 +1,40 @@
...
+                        subscribe(this)
+                        true
+                    }
+
+                override fun onSubscribe(disposable: Disposable) {
+                    emitter.setDisposable(disposable)
+                }
+
+                override fun onComplete() {
+                    if ((count < 0) || (counter!!.addAndGet(-1) >= 0)) {
```
CherryPerry: Maybe we can do it without `count` and `!!`?
`counter == null || counter.addAndGet(-1) >= 0`

PR #311, opened 2019-11-26
Implementing repeatUntil operator

minaEweida:


minaEweida:
This operator is wrong, it was done as a combination of repeat with takeUntil but now it does not make sense since repeat is now implemented.
Still WIP will update shortly

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/RepeatUntil.kt
```diff
@@ -0,0 +1,39 @@
...
+                        true
+                    }
+
+                override fun onSubscribe(disposable: Disposable) {
+                    emitter.setDisposable(disposable)
+                }
+
+                override fun onComplete() {
+                    if (!emitter.isDisposed) {
+                        if (predicate()) {
```
arkivanov: We should try-catch any callback invocations. Write something like:
```
emitter.tryCatch(block = { predicate }) {
    if (it) {
        complete
    } else {
        subscribe
    }
}
```

arkivanov: Also please check `emitter.isDisposed` after predicate invocation (inside tryCatch callback). Because callback might be blocking.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/single/RepeatUntil.kt
```diff
@@ -0,0 +1,6 @@
+package com.badoo.reaktive.single
+
+import com.badoo.reaktive.observable.Observable
+import com.badoo.reaktive.observable.repeatUntil
+
+fun <T> Single<T>.repeatUntil(predicate: () -> Boolean): Observable<T> = asObservable().repeatUntil(predicate)
```
arkivanov: It would be really nice to provide Single result in callback. This is something that is missing in RxJava but would be nice to have. So we will be able to check the result and decide whether we should repeat or not.
So the it should be something like: `repeatUntil(predicate: (T) -> Boolean)`. For `Maybe` is ok as it is now, because `Maybe` may complete without a value and the T is nullable. Looks like a separate implementation would be needed.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/RepeatUntil.kt
arkivanov: And a test for exception in predicate would be nice.

minaEweida: Done

reaktive/src/commonMain/kotlin/com/badoo/reaktive/single/RepeatUntil.kt
minaEweida: Done in a separate commit

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/RepeatUntil.kt
```diff
@@ -0,0 +1,44 @@
...
+
+                override fun onSubscribe(disposable: Disposable) {
+                    emitter.setDisposable(disposable)
+                }
+
+                override fun onComplete() {
+                    emitter.tryCatch(
+                        block = { predicate() },
+                        onSuccess = {
+                            if (!emitter.isDisposed) {
```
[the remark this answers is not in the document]
minaEweida: Will do that, but passing predicate without invoking it will return lambda on success so I will have to do something like 
`emitter.tryCatch(
                        block = { predicate },
                        onSuccess = { shouldRepeat ->
                            if (shouldRepeat()) {
                                emitter.onComplete()
                            } else if (!emitter.isDisposed) {
                                subscribeToUpstream()
                            }
                        }
                    )`
So better send the predicate as { predicate(value) } to get a boolean in onSuccess. Since the lambda expected in block is different than the predicate not like the Observable.repeatUntil

What do you think?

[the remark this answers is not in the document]
arkivanov: tryCatch accepts a function to be invoked. For Observable case predicate has same type so can be passed directly. For Single case the types are different and if you pass directly it won't compile. So it's makes sense only for Observable.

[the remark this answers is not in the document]
minaEweida: Cool, for Observable I am directly passing the predicate but for Single I am passing `{ predicate(value) }`. Check `Singe.RepeatUntil` please.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/single/RepeatUntil.kt
```diff
@@ -0,0 +1,44 @@
...
+                override fun onSubscribe(disposable: Disposable) {
+                    emitter.setDisposable(disposable)
+                }
+
+                override fun onSuccess(value: T) {
+                    emitter.onNext(value)
+
+                    emitter.tryCatch(
+                        block = { predicate(value) },
+                        onSuccess = { shouldRepeat ->
```
arkivanov: The name of `shouldRepeat` should be opposite, e.g. `shouldComplete`. Or just remove and use `it` as with `Observable` case.

minaEweida: sorry my bad

minaEweida: Updated PR

PR #312, opened 2019-11-27
Added mapIterable operators

arkivanov:


reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/MapIterable.kt
```diff
@@ -0,0 +1,6 @@
+package com.badoo.reaktive.observable
+
+fun <T, R> Observable<Iterable<T>>.mapIterable(mapper: (T) -> R): Observable<List<R>> = map { it.map(mapper) }
+
+fun <T, R, C : MutableCollection<R>> Observable<Iterable<T>>.mapIterableTo(collectionSupplier: () -> C, mapper: (T) -> R): Observable<C> =
```
CherryPerry: Do we really need this operator? Why not use just `map { }`?
Also what about KN? You can use here `SharedList`, but it is internal.

arkivanov: I find this operator very useful. E.g. tried to use `map` and here how it looked like:
```
stream
    .map { list ->
        list.map { it.toAnotherObject() }
    }
```
instead we can write like: `stream.mapIterable { it.toAnotherObject() }`

arkivanov: A there any problems you see with K/N? The collection is created and filled on the same thread. Then it might be frozen when passed down the stream. But looks like we don't care about collection's feature after emission.

Issue #313, opened 2019-11-28
Replay count


saket:
It would be great if `replay()` accepted a count of items to replay (from the end) on subscription.

arkivanov:
Accepted.

arkivanov:
WIP

PR #318, opened 2019-12-04
Move ArrayQueue to common

arkivanov:
In order to implement `replay(bufferSize)` operator requested in #313 we need `ArrayQueue` in common source set.

1. Moved `Queue` interface from `JvmJs` to `Common`
2. Moved `ArrayQueue` from `JvmJs` to `Common`
3. Introduced expect class `SharedQueue` in `Common`
4. Implemented actual class `SharedQueue` in `JvmJs` as a typealias to `ArrayQueue`
5. Introduced `CopyOnWriteQueue` in `NativeCommon`
6. Implemented actual class `SharedQueue` in `NativeCommon`: `ArrayQueue` is used if not frozen,  `CopyOnWriteQueue` is used as a fallback when frozen
7. Introduced `MutableFreezableHelper` in `NativeCommon` that helps with switching from non-frozen to frozen variant
8. Introduced `QueueTests` for testing queues: used for `ArrayQueue` and `CopyOnWriteQueue`

I will optimize `SharedList` for K/N same way a bit later (`MutableFreezableHelper` will be reused).


tools/binary-compatibility/reference-public-api/reaktive.txt
```diff
@@ -1238,6 +1238,7 @@ public final class com/badoo/reaktive/utils/queue/ArrayQueue : com/badoo/reaktiv
 	public static final field Companion Lcom/badoo/reaktive/utils/queue/ArrayQueue$Companion;
 	public static final field INITIAL_CAPACITY I
 	public fun <init> ()V
+	public final fun asList ()Ljava/util/List;
```
CherryPerry: Quite strange that ArrayQueue is part of public API. Any suspicions?

CherryPerry: Nevermind, fixed in https://github.com/badoo/Reaktive/pull/319

PR #324, opened 2019-12-07
Added publish with selector operator

### LINKED ISSUE #315: Equivalent of RxJava's ObservablePublishSelector?
RxJava has a nifty operator variant of `publish(Function)` that offers a scoped way to consume published events, removing the requirement of manually calling `connect()`. 

```kotlin
Observable.just("value")
  .publish { publishedEvents -> 
    // connect() is called when merge() subscribes.
    merge(consumerA(publishedEvents), consumerB(publishedEvents)
  }
```

Is it possible to write a similar operator by using the existing operators in Reaktive? Or would that require a custom operator?

arkivanov:
Fixes #315


reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/PublishWithSelector.kt
```diff
@@ -0,0 +1,16 @@
+package com.badoo.reaktive.observable
```
CherryPerry: Let's move it to `Publish.kt` file? It is small, has same name, so it will be easier to find it. Also we won't generate additional class for JVM.

arkivanov:
Sorry, had to fix tests compilation

PR #328, opened 2019-12-10
Freeze tasks in iOS MainScheduler if needed

### LINKED ISSUE #327: Callbacks must be frozen when passed to iOS main queue
After replacing `AtomicReference` with `FreezableAtomicReference` callbacks are not frozen automatically anymore. See `MainScheduler` class for iOS.

I think we should check if current thread is NOT main, then freeze the whole `Operation`. Not freezing the callbacks leads to unpredictable crashes.

PS: Also we can replace Reaktive's `AtomicReference` with K/N variant.

arkivanov:
Fixes #327


arkivanov:
@CherryPerry could you please run iOS sample app?

CherryPerry:
Works fine.

PR #331, opened 2019-12-13
Implement convenient extension to the zip operator.

mk-jones:
Project reactor has a very convenient zipWith operator. Thought reaktive could benefit from it too.

https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Mono.html#zipWith-reactor.core.publisher.Mono-

### DIFF
--- a/reaktive/src/commonMain/kotlin/com/badoo/reaktive/maybe/ZipWith.kt
+++ b/reaktive/src/commonMain/kotlin/com/badoo/reaktive/maybe/ZipWith.kt
@@ -2,6 +2,3 @@ package com.badoo.reaktive.maybe
 
 fun <T, R> Maybe<T>.zipWith(maybe: Maybe<R>): Maybe<Pair<T, R>> =
     zip(this, maybe) { first, second -> first to second }
-
-fun <T, R, I> Maybe<Pair<T, R>>.zipWith(maybe: Maybe<I>): Maybe<Triple<T, R, I>> =
-    zip(this, maybe) { pair, other -> Triple(pair.first, pair.second, other) }
--- a/reaktive/src/commonMain/kotlin/com/badoo/reaktive/maybe/ZipWith.kt
+++ b/reaktive/src/commonMain/kotlin/com/badoo/reaktive/maybe/ZipWith.kt
@@ -1,7 +1,7 @@
 package com.badoo.reaktive.maybe
 
-fun <T, R> Maybe<T>.zipWith(maybe: Maybe<R>): Maybe<Pair<T, R>> =
-    zip(this, maybe) { first, second -> first to second }
+fun <T, R> Maybe<T>.zipWith(other: Maybe<R>): Maybe<Pair<T, R>> =
+    zip(this, other) { first, second -> first to second }
 
-fun <T, R, I> Maybe<T>.zipWith(maybe: Maybe<R>, mapper: (T, R) -> I): Maybe<I> =
-    zip(this, maybe) { first, second -> mapper(first, second) }
+fun <T, R, I> Maybe<T>.zipWith(other: Maybe<R>, mapper: (T, R) -> I): Maybe<I> =
+    zip(this, other) { first, second -> mapper(first, second) }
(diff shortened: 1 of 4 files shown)


reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/ZipWith.kt
```diff
@@ -0,0 +1,4 @@
+package com.badoo.reaktive.observable
+
+fun <T, R> Observable<T>.zipWith(observable: Observable<R>): Observable<Pair<T, R>> =
```
arkivanov: It would be good to provide a zipper function, so the signature should be:
```
fun <T, S, R> Observable<T>.zipWith(other: Observable<S>, zipper: (T, S) -> R): Observable<R> = 
    zip(this, other, zipper)
```

And same for `Maybe` and `Single`.

arkivanov: Also please rename `observable` to `other` to keep consistency with the reset of the code.

mk-jones: Added the mappers for Observable, Single and Maybe. 
Also renamed the other stream to "other" in all of them.

mk-jones: Added a variant with a mapper to the zipWith method.
Also renamed the other stream to "other" in all of them.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/maybe/ZipWith.kt
```diff
@@ -0,0 +1,7 @@
+package com.badoo.reaktive.maybe
+
+fun <T, R> Maybe<T>.zipWith(other: Maybe<R>): Maybe<Pair<T, R>> =
```
arkivanov: Do you think we really need this variant in the library? You could just write: `a.zipWith(b, ::Pair)`.

mk-jones: I've been thinking about it.
So I think having zipWith(other) is useful in that it encourages people to use it without a zipper function. That is, it highlights that you can just combine two requests as a pair. It also reads slightly nicer.
```suggestion
fun loadSomething() =
    dataSource.loadSome()
        .zipWith(dataSource.loadThing())

fun loadSomething() =
    dataSource.loadSome()
        .zipWith(dataSource.loadThing(), ::Pair)
```
However, this is so small I don't think it really matters either way. I'll let you decide.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/maybe/ZipWith.kt
```diff
@@ -0,0 +1,7 @@
...
+    zip(this, other) { first, second -> first to second }
+
+fun <T, R, I> Maybe<T>.zipWith(other: Maybe<R>, mapper: (T, R) -> I): Maybe<I> =
```
mk-jones: mapper should be changed to zipper

reaktive/src/commonMain/kotlin/com/badoo/reaktive/maybe/ZipWith.kt
arkivanov: Cool, then I don't think we should have such a simple shortcuts in the library :-)

reaktive/src/commonMain/kotlin/com/badoo/reaktive/maybe/ZipWith.kt
arkivanov: `zip(a, b, mapper)` has the lambdas named as `mapper` actually, `combineLatest` as well. This is small but maybe it's better to keep consistency?

mk-jones: Sure

arkivanov:
Please re-run the API test

PR #339, opened 2019-12-14
Upgrade Detekt version to 1.2.2 and merge detekt.yml with default

CherryPerry:


detekt.yml
```diff
@@ -539,7 +552,7 @@ style:
   UnusedImports:
     active: false
   UnusedPrivateClass:
-    active: false
```
arkivanov: Why do we enable all these checks now?

arkivanov:
Also is there a check for `switch` replaceable with `if-else`?

detekt.yml
CherryPerry: Because they were enabled in https://github.com/arturbosch/detekt/blob/master/detekt-cli/src/main/resources/default-detekt-config.yml
I merged our config with default config, so all new checks are imported correctly.

CherryPerry:
https://github.com/arturbosch/detekt/blob/master/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseIfInsteadOfWhen.kt

arkivanov:
> https://github.com/arturbosch/detekt/blob/master/detekt-rules/src/main/kotlin/io/gitlab/arturbosch/detekt/rules/style/UseIfInsteadOfWhen.kt

Let's enable?

CherryPerry:
Done

PR #346, opened 2019-12-20
Added timeout operator

### LINKED ISSUE #330: Timeout Rx operator
Hi, does reaktive have a timeout Rx operator?

http://reactivex.io/documentation/operators/timeout.html

arkivanov:
Fixes #330


reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/Timeout.kt
```diff
@@ -0,0 +1,42 @@
+package com.badoo.reaktive.observable
+
+import com.badoo.reaktive.base.CompositeDisposableObserver
+import com.badoo.reaktive.base.exceptions.TimeoutException
+import com.badoo.reaktive.completable.CompletableCallbacks
+import com.badoo.reaktive.disposable.Disposable
+import com.badoo.reaktive.disposable.addTo
+import com.badoo.reaktive.scheduler.Scheduler
+
+fun <T> Observable<T>.timeout(timeoutMillis: Long, scheduler: Scheduler, other: Observable<T>? = null): Observable<T> =
```
CherryPerry: `scheduler: Scheduler = computationScheduler`?

arkivanov: At the moment we do not define default values for scheduler arguments. E.g. debounce  or sample. I find it error prone, since you may expect emissions to be on main thread.

mk-jones:
@arkivanov @CherryPerry 
So if I understood the code correctly, the timeout functionality only starts after the first item is emitted. In the RxJava3 and Reactor implementations it is started in the subscribe flow:
 - https://github.com/ReactiveX/RxJava/blob/a1693ecc9215027a7a3eae75272979ebb4d79a27/src/main/java/io/reactivex/rxjava3/internal/operators/observable/ObservableTimeoutTimed.java
 - https://github.com/reactor/reactor-core/blob/c6b75f2578ee5f201ac06add2239322cc5b7bfcd/reactor-core/src/main/java/reactor/core/publisher/FluxTimeout.java

Is this intended? I think we should be matching implementation, as that variant seems to be more useful.
It would also be useful to have this implemented for Single and Maybe as well.

arkivanov:
Hey, thanks for reporting the bug. Will fix!

arkivanov:
Created #354

PR #360, opened 2019-12-27
Move ThreadLocalStorage to utils module

### LINKED ISSUE #343: ThreadLocalStorage should be in "utils" module

arkivanov:
Fixes #343


reaktive/src/commonMain/kotlin/com/badoo/reaktive/utils/ThreadLocalStorage.kt
```diff
@@ -1,80 +1,10 @@
...
-            }
-        }
-    }
-}
+@Deprecated(
+    message = "Use ThreadLocalHolder from 'utils' package. This class will be removed soon.",
+    replaceWith = ReplaceWith("ThreadLocalHolder", "com.badoo.reaktive.utils.ThreadLocalHolder"),
+    level = DeprecationLevel.WARNING
+)
+open class ThreadLocalStorage<T : Any>(initialValue: T? = null) : ThreadLocalHolder<T>(initialValue), Disposable
```
CherryPerry: Why `open`?

arkivanov: Good point, this can be still final. Anyway it's deprecated now, don't think it's worth changing. Ok?

CherryPerry: Ok

Issue #350, opened 2019-12-21
Parallelise build on macOS


CherryPerry:
After adding more targets for macOS https://github.com/badoo/Reaktive/pull/349, build time increased in 3 times.
We should introduce new `MP_TARGET` and add more virtual machines, to build different darwin targets in parallel.

Issue #345, opened 2019-12-20
Binary compatibility should not download native toolchain


CherryPerry:
We can save some time on check if we disable native targets for compatibility checks.
Probably we need to introduce new `MP_TARGET` that defines JVM only behaviour.

See [build](https://github.com/badoo/Reaktive/runs/358045328).

PR #365, opened 2019-12-29
Introduce new targets and refactor Target

CherryPerry:
Fixes https://github.com/badoo/Reaktive/issues/350 
Fixes https://github.com/badoo/Reaktive/issues/345

- ~~Introduce new targets~~
- ~~Split mac os hosted targets~~
- ~~Split linux hosted targets~~


settings.gradle.kts
```diff
@@ -0,0 +1,21 @@
+import org.gradle.api.internal.FeaturePreviews.Feature
```
CherryPerry: Revert, do not use internal api.

CherryPerry:
Back to 20 minutes per commit!

PR #377, opened 2020-01-08
Annotate DisposableScope as experimental

arkivanov:


reaktive-annotations/src/commonMain/kotlin/com/badoo/reaktive/annotations/Annotations.kt
```diff
@@ -41,3 +41,7 @@ annotation class EventsOnComputationScheduler
 @Retention(AnnotationRetention.SOURCE)
 @Target(AnnotationTarget.FUNCTION, AnnotationTarget.FIELD, AnnotationTarget.PROPERTY)
 annotation class EventsOnIoScheduler
+
+@Experimental(Experimental.Level.WARNING)
+@Retention(AnnotationRetention.BINARY)
+annotation class ExperimentalReaktiveApi
```
CherryPerry: I think you should move this annotation to main package.
Experimental annotation is totally unrelated to the content of this package.

Issue #381, opened 2020-01-11
Bintray publications are broken


arkivanov:
Version 1.1.7 metadata contains only Darwin targets, no Android, JVM, Linux: https://dl.bintray.com/badoo/maven/com/badoo/reaktive/reaktive/1.1.7/

We need to fix this and re-publish

PR #382, opened 2020-01-12
Handle META target in more precise way when publishing

CherryPerry:
Fixes https://github.com/badoo/Reaktive/issues/381

### DIFF
--- a/buildSrc/src/main/kotlin/PublishPlugin.kt
+++ b/buildSrc/src/main/kotlin/PublishPlugin.kt
@@ -33,23 +33,23 @@ abstract class PublishPlugin : Plugin<Project> {
 
     private fun createConfigurationMap(project: Project): Map<String, Boolean> {
         return mapOf(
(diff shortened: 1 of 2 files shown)


buildSrc/src/main/kotlin/PublishPlugin.kt
```diff
@@ -93,6 +93,7 @@ abstract class PublishPlugin : Plugin<Project> {
         target.tasks.named(BintrayUploadTask.getTASK_NAME(), BintrayUploadTask::class) {
             dependsOn(project.tasks.named(TASK_FILTERED_PUBLISH_TO_MAVEN_LOCAL))
             doFirst {
+                logger.warn("Publication configuration map: $taskConfigurationMap")
```
arkivanov: Is it ok to keep?

CherryPerry: Yes, logs will be more clear.

Issue #379, opened 2020-01-11
Mark classes that use DisposableScope with UseExperimental


CherryPerry:
```
w: /Users/user/IdeaProjects/Reaktive/sample-mpp-module/src/commonMain/kotlin/com/badoo/reaktive/samplemppmodule/store/KittenStoreImpl.kt: (16, 37): This declaration is experimental and its usage should be marked with '@com.badoo.reaktive.annotations.ExperimentalReaktiveApi' or '@UseExperimental(com.badoo.reaktive.annotations.ExperimentalReaktiveApi::class)'
```

mayokunthefirst:
Hello @CherryPerry would love to get on this.

CherryPerry:
Sure, feel free.

CherryPerry:
Not all of them were fixed in #386

```
w: /Users/runner/runners/2.163.1/work/Reaktive/Reaktive/sample-mpp-module/src/commonMain/kotlin/com/badoo/reaktive/samplemppmodule/binder/KittenBinder.kt: (13, 33): This declaration is experimental and its usage should be marked with '@com.badoo.reaktive.annotations.ExperimentalReaktiveApi' or '@UseExperimental(com.badoo.reaktive.annotations.ExperimentalReaktiveApi::class)'
w: /Users/runner/runners/2.163.1/work/Reaktive/Reaktive/sample-mpp-module/src/commonMain/kotlin/com/badoo/reaktive/samplemppmodule/binder/KittenBinder.kt: (22, 26): This declaration is experimental and its usage should be marked with '@com.badoo.reaktive.annotations.ExperimentalReaktiveApi' or '@UseExperimental(com.badoo.reaktive.annotations.ExperimentalReaktiveApi::class)'
> Task :sample-mpp-module:compileKotlinMacosX64
w: /Users/runner/runners/2.163.1/work/Reaktive/Reaktive/sample-mpp-module/src/commonMain/kotlin/com/badoo/reaktive/samplemppmodule/binder/KittenBinder.kt: (25, 17): This declaration is experimental and its usage should be marked with '@com.badoo.reaktive.annotations.ExperimentalReaktiveApi' or '@UseExperimental(com.badoo.reaktive.annotations.ExperimentalReaktiveApi::class)'
[...]

PR #387, opened 2020-01-14
Mark classes that use DisposableScope with UseExperimental

mayokunthefirst:
Fixes issue #379

### DIFF
--- a/sample-mpp-module/src/commonMain/kotlin/com/badoo/reaktive/samplemppmodule/binder/KittenBinder.kt
+++ b/sample-mpp-module/src/commonMain/kotlin/com/badoo/reaktive/samplemppmodule/binder/KittenBinder.kt
@@ -1,5 +1,6 @@
 package com.badoo.reaktive.samplemppmodule.binder
 
+
 import com.badoo.reaktive.annotations.ExperimentalReaktiveApi
 import com.badoo.reaktive.disposable.scope.DisposableScope
 import com.badoo.reaktive.disposable.scope.disposableScope
@@ -7,7 +8,7 @@ import com.badoo.reaktive.observable.map
 import com.badoo.reaktive.samplemppmodule.store.KittenStoreBuilder
 import com.badoo.reaktive.samplemppmodule.view.KittenView
 
-@ExperimentalReaktiveApi
+@UseExperimental(ExperimentalReaktiveApi::class)
 class KittenBinder(
     storeBuilder: KittenStoreBuilder
 ) {
--- a/sample-mpp-module/src/commonMain/kotlin/com/badoo/reaktive/samplemppmodule/binder/KittenBinder.kt
+++ b/sample-mpp-module/src/commonMain/kotlin/com/badoo/reaktive/samplemppmodule/binder/KittenBinder.kt
@@ -1,11 +1,13 @@
 package com.badoo.reaktive.samplemppmodule.binder
 
+import com.badoo.reaktive.annotations.ExperimentalReaktiveApi
 import com.badoo.reaktive.disposable.scope.DisposableScope
 import com.badoo.reaktive.disposable.scope.disposableScope
 import com.badoo.reaktive.observable.map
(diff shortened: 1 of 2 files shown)


CherryPerry:
```
w: /Users/runner/runners/2.163.1/work/Reaktive/Reaktive/sample-mpp-module/src/darwinCommonMain/kotlin/com/badoo/reaktive/samplemppmodule/KittenStoreBuilderImpl.kt: (9, 41): This declaration is experimental and its usage should be marked with '@com.badoo.reaktive.annotations.ExperimentalReaktiveApi' or '@UseExperimental(com.badoo.reaktive.annotations.ExperimentalReaktiveApi::class)'
```

Now `KittenStoreBuilderImpl` uses experimental features.

I think it is better to use `@UseExperimental(ExperimentalReaktiveApi::class)` in `sample-mpp-module` for all cases, because module itself just uses new unstable API and do not exposes it.

So my proposal – mark all files that uses `DisposableScope` with `@UseExperimental(ExperimentalReaktiveApi::class)` and it should be enough. But better to check it locally and on CI after change.

mayokunthefirst:
> ```
> w: /Users/runner/runners/2.163.1/work/Reaktive/Reaktive/sample-mpp-module/src/darwinCommonMain/kotlin/com/badoo/reaktive/samplemppmodule/KittenStoreBuilderImpl.kt: (9, 41): This declaration is experimental and its usage should be marked with '@com.badoo.reaktive.annotations.ExperimentalReaktiveApi' or '@UseExperimental(com.badoo.reaktive.annotations.ExperimentalReaktiveApi::class)'
> ```
> 
> Now `KittenStoreBuilderImpl` uses experimental features.
> 
> I think it is better to use `@UseExperimental(ExperimentalReaktiveApi::class)` in `sample-mpp-module` for all cases, because module itself just uses new unstable API and do not exposes it.
> 
> So my proposal – mark all files that uses `DisposableScope` with `@UseExperimental(ExperimentalReaktiveApi::class)` and it should be enough. But better to check it locally and on CI after change.

@CherryPerry after implementing this, I keep getting these warnings:
````
w: /home/runner/work/Reaktive/Reaktive/reaktive/src/commonMain/kotlin/com/badoo/reaktive/annotations/Annotations.kt: (5, 2): This class can only be used with the compiler argument '-Xuse-experimental=kotlin.Experimental'
w: /home/runner/work/Reaktive/Reaktive/reaktive/src/commonMain/kotlin/com/badoo/reaktive/annotations/Annotations.kt: (5, 15): This class can only be used with the compiler argument '-Xuse-experimental=kotlin.Experimental'
w: /home/runner/work/Reaktive/Reaktive/reaktive/src/commonMain/kotlin/com/badoo/reaktive/disposable/scope/DisposableScopeImpl.kt: (16, 2): This class can only be used with the compiler argument '-Xuse-experimental=kotlin.Experimental'
````

CherryPerry:
@mayokunthefirst It is fine.

mayokunthefirst:
> @mayokunthefirst It is fine.

@CherryPerry Oh, awesome then, but out of curiosity, if it's not too much too ask, would love to know if getting rid of such warning would involve something like adding such line `-Xuse-experimental=kotlin.Experimental` to the `gradle.properties` file, something similar to enabling incremental annotation processing `kapt.incremental.apt` or enabling jetifier i.e. `android.enableJetifier`?

CherryPerry:
It is a little harder to do. You need to add special parameter to every source set in every module. I will think, how to implement it better than just applying it via `allProjects { ... }`.

[Link.](https://kotlinlang.org/docs/reference/experimental.html#module-wide-use)

mayokunthefirst:
> It is a little harder to do. You need to add special parameter to every source set in every module. I will think, how to implement it better than just applying it via `allProjects { ... }`.
> 
> [Link.](https://kotlinlang.org/docs/reference/experimental.html#module-wide-use)

Alright then, thank you.

PR #393, opened 2020-01-31
Add wrappers for Swift

### LINKED ISSUE #368: Investigate Swift extensions
Reaktive can be exposed to Swift because it is not based on coroutines. But calling its operators does not look nice. We need to investigate if we can provide Swift extensions.

arkivanov:
As I wrote in #368 generics for interfaces are not supported in Objective-C. So looks like there is no way to write good Swift extensions for Reaktive interfaces with generics. I came up with a workaround: wrapper classes.

Example:
```
class MyExposedDataSource {
    fun load(): SingleWrapper<String> =
        singleFromFunction { "Loaded data" }
            .subscribeOn(ioScheduler)
            .observeOn(mainScheduler)
            .wrap()
}
```

The returned `Single` still can be used in a normal way from Kotlin and at the same time can be exposed to Swift with generics.

Fixes #368


CherryPerry:
Do you want to add it to readme section? Or maybe use in samples?

arkivanov:
> Do you want to add it to readme section? Or maybe use in samples?

Could not find a good use case for a shared stream in sample. If you have any ideas let me know.
Updated readme. Also fixed wrappers' subscribe methods.

dcvz:
@arkivanov is there something similar for `BehaviorSubject`. I'd like to be able to use the types with Swift while still being able to publish new values.

arkivanov:
> @arkivanov is there something similar for `BehaviorSubject`? I'd like to be able to use the types with Swift while still being able to publish new values.

No, there nothing like this provided by the library. But you can create your own wrapper:
```kotlin
import com.badoo.reaktive.observable.subscribe as subscribeRx

class BehaviorSubjectWrapper<T>(initialValue: T) : BehaviorSubject<T> by BehaviorSubject(initialValue) {
    @UseReturnValue
    fun subscribe(
        isThreadLocal: Boolean = false,
        onSubscribe: ((Disposable) -> Unit)? = null,
        onError: ((Throwable) -> Unit)? = null,
        onComplete: (() -> Unit)? = null,
        onNext: ((T) -> Unit)? = null
    ): Disposable =
        subscribeRx(
            isThreadLocal = isThreadLocal,
            onSubscribe = onSubscribe,
            onError = onError,
            onComplete = onComplete,
            onNext = onNext
        )
}
```

PR #395, opened 2020-02-01
Add deferred builders

arkivanov:


reaktive/src/commonMain/kotlin/com/badoo/reaktive/completable/Various.kt
```diff
@@ -41,3 +41,14 @@ fun completableFromFunction(func: () -> Unit): Completable =
         func()
         emitter.onComplete()
     }
+
+fun completableDeferred(supplier: () -> Completable): Completable =
```
CherryPerry: Maybe put in separate file?

arkivanov: Ok. What do you think of naming? `observableDeferred` or `observableDefer`?

CherryPerry: Both fine for me.
But probably `defer` is [better](http://reactivex.io/documentation/operators/defer.html).

PR #430, opened 2020-03-13
Update replay after rxjava bug fix

### LINKED ISSUE #420: Inconsistency in behavior of replay
Hey.

I wrote a test and run it on both RxJava and Reaktive and observed different results. Here is the test:

```
@Test
fun testReplayRefCount() {
    val subject = BehaviorSubject.createDefault(false)
    val sharedStream = subject.replay(1).refCount(1)

    var disposable = sharedStream.test()
    subject.onNext(true)

    disposable.assertValues(false, true)
    disposable.dispose()

    disposable = sharedStream.test()
    subject.onNext(false)

    disposable.assertValues(true, false)
}
```

As you might guess, running this test with RxJava results in green, while with Reaktive - red.
Looks like on second subscription, the internal implementation of replay creates a new [ReplaySubject](https://github.com/badoo/Reaktive/blob/master/reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/Replay.kt#L11) which results in a situation where all subsequent `onNext` calls are happen for newly created instance and lost for an existing one.

arkivanov:
So there was a bug in RxJava's `Observable.replay()` operator, and it was derived by Reaktive. This bug was recently fixed (https://github.com/ReactiveX/RxJava/issues/6920). We need to update our implementation as well.

This also fixes #420.

### DIFF
--- a/reaktive/src/commonMain/kotlin/com/badoo/reaktive/base/operator/Publish.kt
+++ b/reaktive/src/commonMain/kotlin/com/badoo/reaktive/base/operator/Publish.kt
@@ -64,8 +64,13 @@ internal fun <T> Observable<T>.publish(subjectFactory: () -> Subject<T>): Connec
 
         override fun subscribe(observer: ObservableObserver<T>) {
             state
-                .updateAndGet {
-                    it ?: PublishState.NotConnected(subject = subjectFactory())
+                .updateAndGet<PublishState<T>?, PublishState<T>> {
+                    when (it) {
+                        is PublishState.NotConnected,
+                        is PublishState.Connected -> it
+                        is PublishState.Disconnected,
+                        null -> PublishState.NotConnected(subject = subjectFactory())
+                    }
                 }
                 .subject
                 .subscribe(observer)
--- a/reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/ReplayTest.kt
+++ b/reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/ReplayTest.kt
@@ -28,16 +28,16 @@ class ReplayTest : PublishGenericTests by PublishGenericTestsImpl({ replay() }) 
     }
 
     @Test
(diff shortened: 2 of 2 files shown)


sample-android-app/src/main/java/com/badoo/reaktive/sample/android/ReaktiveAcitivy.kt
```diff
@@ -0,0 +1,54 @@
...
+import com.badoo.reaktive.observable.subscribeOn
+import com.badoo.reaktive.scheduler.computationScheduler
+import com.badoo.reaktive.scheduler.ioScheduler
+import com.badoo.reaktive.single.singleFromFunction
+import com.badoo.reaktive.single.subscribeOn
+
+    @SuppressLint("Registered")
+
+
+    class ReaktiveActivity : AppCompatActivity() {
```
CherryPerry: What is it?
Why it is in this PR?
What's up with formatting?

sample-android-app/src/test/java/com/ReaktiveTests.kt
```diff
@@ -0,0 +1,127 @@
...
+        subject.onNext(0)
+        observer.dispose()
+        disposable.dispose()
+
+        val observer2 = sharedStream.test()
+        sharedStream.connect()
+        observer2.assertValue(0)
+    }
+
+//    @Test
```
CherryPerry: ?

sample-android-app/src/main/java/com/badoo/reaktive/sample/android/ReaktiveAcitivy.kt
arkivanov: By accident, removed

sample-android-app/src/test/java/com/ReaktiveTests.kt
arkivanov: Removed, sorry

Issue #445, opened 2020-03-30
Change collections visibility


vchernyshov:
Change visibility of collection from package `com.badoo.reaktive.utils.queue` to public it will allow  make thread safe custom operators for iOS target.

arkivanov:
Thanks for raising the issue. Here are some explanations. We made things internal for a reason: we need flexibility without any binary compatibility obligations. Queue APIs in particular are changing from time to time, they even were refactored once. If we make them public then we will not be able to change.

Don't hesitate to copy-paste any code you need. Just use different package name so there will be no clashes.

vchernyshov:
Ok, I got it) I will continue copy-paste approach))

PR #446, opened 2020-03-30
Make Queue public

vchernyshov:
Changes due to #445 issue.


arkivanov:
Please see comment the comment in #445

PR #453, opened 2020-05-01
Non-null wrappers and additional overloads

### LINKED ISSUE #452: Provide a non-null type constraint for better interop with swift
Due to constraints related to how obj-c and swift handle nullable types, generic type parameters from kotlin get converted into nullable types unless they have a non-null constraint:

image

Reaktive suffers from the same problem: its value callbacks receive a nullable types:

image

Will it be possible to make this change in `ValueCallback`?

```kotlin
interface ValueCallback<in T : Any> {
```

arkivanov:
#Fixes #452

### DIFF
--- a/reaktive/src/commonMain/kotlin/com/badoo/reaktive/completable/CompletableWrapper.kt
+++ b/reaktive/src/commonMain/kotlin/com/badoo/reaktive/completable/CompletableWrapper.kt
@@ -28,6 +28,9 @@ open class CompletableWrapper(inner: Completable) : Completable by inner {
         )
 
     @UseReturnValue
+    fun subscribe(): Disposable = subscribeRx()
+
+    @UseReturnValue
     fun subscribe(
         isThreadLocal: Boolean = false,
         onComplete: () -> Unit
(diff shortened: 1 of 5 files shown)


reaktive/src/commonMain/kotlin/com/badoo/reaktive/completable/CompletableWrapper.kt
```diff
@@ -26,6 +26,28 @@ open class CompletableWrapper(inner: Completable) : Completable by inner {
             onError = onError,
             onComplete = onComplete
         )
+
+    @UseReturnValue
+    fun subscribe(
```
mk-jones, line 31: I think it would be nice to include the subscribe option with no params passed.

arkivanov: Good catch!

arkivanov: Fixed

PR #460, opened 2020-05-22
Extract read only interfaces for BehaviorSubject

### LINKED ISSUE #456: Extract read-only interface for BehaviorSubject
Currently there is no way to represent the `BehaviorSubject` as `Observable` with current `value` property and without inputs. This might be useful for property-like cases when we want to expose a read-only `BehaviorSubject`. Or just for better SRP when we only need reading.

arkivanov:
Fixes #456 

Extracted `BehaviorRelay` and `BehaviorObservable` interfaces. Please check the binary compatibility files change as well, looks like it should not break anything.

### DIFF
--- a/reaktive/src/commonMain/kotlin/com/badoo/reaktive/subject/behavior/BehaviorObservable.kt
+++ b/reaktive/src/commonMain/kotlin/com/badoo/reaktive/subject/behavior/BehaviorObservable.kt
@@ -1,0 +1,8 @@
+package com.badoo.reaktive.subject.behavior
+
+import com.badoo.reaktive.observable.Observable
+
+interface BehaviorObservable<out T> : Observable<T> {
+
+    val value: T
+}
--- a/tools/binary-compatibility/reference-public-api/reaktive.txt
+++ b/tools/binary-compatibility/reference-public-api/reaktive.txt
@@ -1249,8 +1249,14 @@ public final class com/badoo/reaktive/subject/SubjectExtKt {
 	public static final fun isActive (Lcom/badoo/reaktive/subject/Subject;)Z
 }
 
-public abstract interface class com/badoo/reaktive/subject/behavior/BehaviorSubject : com/badoo/reaktive/subject/Subject {
+public abstract interface class com/badoo/reaktive/subject/behavior/BehaviorObservable : com/badoo/reaktive/observable/Observable {
 	public abstract fun getValue ()Ljava/lang/Object;
+}
+
+public abstract interface class com/badoo/reaktive/subject/behavior/BehaviorRelay : com/badoo/reaktive/subject/Relay, com/badoo/reaktive/subject/behavior/BehaviorObservable {
+}
+
+public abstract interface class com/badoo/reaktive/subject/behavior/BehaviorSubject : com/badoo/reaktive/subject/Subject, com/badoo/reaktive/subject/behavior/BehaviorRelay {
 }
 
 public final class com/badoo/reaktive/subject/behavior/BehaviorSubjectBuilderKt {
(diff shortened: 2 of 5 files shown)


tools/binary-compatibility/reference-public-api/reaktive.txt
```diff
@@ -1225,10 +1225,16 @@ public final class com/badoo/reaktive/subject/SubjectExtKt {
...
 
-public abstract interface class com/badoo/reaktive/subject/behavior/BehaviorSubject : com/badoo/reaktive/subject/Subject {
+public abstract interface class com/badoo/reaktive/subject/behavior/BehaviorObservable : com/badoo/reaktive/observable/Observable {
 	public abstract fun getValue ()Ljava/lang/Object;
 }
 
+public abstract interface class com/badoo/reaktive/subject/behavior/BehaviorRelay : com/badoo/reaktive/subject/Relay, com/badoo/reaktive/subject/behavior/BehaviorObservable {
+}
+
+public abstract interface class com/badoo/reaktive/subject/behavior/BehaviorSubject : com/badoo/reaktive/subject/Subject, com/badoo/reaktive/subject/behavior/BehaviorRelay {
```
CherryPerry, line 1235: I think it should be fine.
https://wiki.eclipse.org/Evolving_Java-based_APIs_2#Achieving_API_Binary_Compatibility
Move API method up type hierarchy | If method in supertype need not be reimplemented by Client | Binary compatible
-- | -- | --

arkivanov: Do you like the change in general?

CherryPerry: Do not have any objections.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/subject/behavior/BehaviorObservable.kt
```diff
@@ -0,0 +1,8 @@
+package com.badoo.reaktive.subject.behavior
+
+import com.badoo.reaktive.observable.Observable
+
+interface BehaviorObservable<out T> : Observable<T> {
+
+    val value: T
```
lukaville, line 7: should we further segregate this interface and create something like value provider and compose BehaviourObservable from Observable and value provider?

arkivanov: Makes sense! Thanks! @mk-jones, @cherry-perry what do you think?

CherryPerry: Not sure.
When we need value provider, we use just kotlin lamda most of the time.

arkivanov: Fare point)

arkivanov: A would still avoid extracting the `value provider` interface at this point, but we can do it at any time later.

arkivanov: @mk-jones What is your opinion?

mk-jones: I don't yet see a use-case for the value provider interface. I wouldn't extract it without having a use-case in mind.

arkivanov: Thanks!

PR #463, opened 2020-05-29
Add amb operators

arkivanov:
Closes: #405 

Doc: http://reactivex.io/documentation/operators/amb.html

### DIFF
--- a/reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/Amb.kt
+++ b/reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/Amb.kt
@@ -1,0 +1,59 @@
+package com.badoo.reaktive.observable
+
+import com.badoo.reaktive.disposable.CompositeDisposable
+import com.badoo.reaktive.disposable.Disposable
+import com.badoo.reaktive.disposable.minusAssign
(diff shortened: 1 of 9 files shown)


reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/Amb.kt
```diff
@@ -0,0 +1,59 @@
...
+import com.badoo.reaktive.disposable.Disposable
+import com.badoo.reaktive.disposable.minusAssign
+import com.badoo.reaktive.disposable.plusAssign
+import com.badoo.reaktive.utils.ObjectReference
+
+fun <T> Iterable<Observable<T>>.amb(): Observable<T> =
+    observable { emitter ->
+        val disposables = CompositeDisposable()
+        emitter.setDisposable(disposables)
+        val markerDisposable = Disposable()
```
CherryPerry, line 13: Why do we need `markerDisposable`? Simple `AtomicBoolean` won't work?

arkivanov: Good catch, fixed

PR #464, opened 2020-05-29
Print stack trace in assertNotError

### LINKED ISSUE #462: Print error stacktrace when TestObserver.assertNotError() fails
Printing `"Source failed"` is not very useful when `test().assertNotError()` is used in a test. What do you think of printing the error stacktrace as well?

arkivanov:
Fixes #462

### DIFF
--- a/reaktive-testing/src/commonMain/kotlin/com/badoo/reaktive/test/base/TestObserverExt.kt
+++ b/reaktive-testing/src/commonMain/kotlin/com/badoo/reaktive/test/base/TestObserverExt.kt
@@ -1,27 +1,50 @@
 package com.badoo.reaktive.test.base
 
+import com.badoo.reaktive.utils.printStack
 import kotlin.test.assertEquals
 import kotlin.test.assertFalse
 import kotlin.test.assertNotNull
 import kotlin.test.assertNull
 import kotlin.test.assertTrue
+import kotlin.test.fail
 
 fun TestObserver.assertError() {
     assertTrue(isError, "Source did not fail")
 }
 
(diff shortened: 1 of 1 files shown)


reaktive-testing/src/commonMain/kotlin/com/badoo/reaktive/test/base/TestObserverExt.kt
```diff
@@ -1,27 +1,50 @@
...
+import kotlin.test.fail
 
 fun TestObserver.assertError() {
     assertTrue(isError, "Source did not fail")
 }
 
 fun TestObserver.assertError(expectedError: Throwable) {
-    assertEquals(expectedError, error, "Source error does not match")
+    val error = error
+    if (error == null) {
```
CherryPerry, line 17: You can beautify it a little. But up to you, `when` can be overkill here and I don't know, will it be optimised by compiler.

```kotlin
when (val error = error) {
     null -> fail()
     else -> { ... }
}
```

arkivanov: Personally I don't like replacing `if-else` with `when` :-)

arkivanov: Personally I don't like replacing `if-else` with `when`, let's keep it like this?

CherryPerry: Keep it.

Issue #443, opened 2020-03-27
Flaky CI "Set up cache" step


arkivanov:
The step "Set up cache" is flaky. If it fails then rerunning it will fail all jobs.

Latest example: https://github.com/badoo/Reaktive/runs/539845234

PR #466, opened 2020-05-29
cache@v2

CherryPerry:
Can fix potentially https://github.com/badoo/Reaktive/issues/443


arkivanov:
Looks like there the build stuck

PR #470, opened 2020-05-30
Fix TestScheduler advanceTime

minaEweida:
Advancing time does not account for resubscriptions for example in `retry` or `repeat` when it's conditional
Reason is the scheduler will just advance time and check if any current tasks are scheduled, it will just execute

But in case any of these tasks will cause a resubscription that should run inside the time frame of the advanced time

For example,
If we have a repeatUntil that will delay 10 seconds for 5 times (based on a network call that will return the number of repeats and interval)

If we advance time by 50 seconds, it is expected to have 5 emissions but that's not the case, we will just get 1.

Proposed fix:
Applying what was suggested in https://github.com/badoo/Reaktive/pull/470#issuecomment-636376085
Instead of blindly advancing time, it will advance time and update task list till no more tasks are expected to be executed

Try running this test which passes using RxJava's TestScheduler, it will fail with Reaktive's TestScheduler.

```    
    private val repeatCounter = AtomicInt(0)

    private fun getRepeatObservable(repeats: Int): Single<Repeater> =
        singleFromFunction {
            val repeatCount = repeatCounter.addAndGet(1)
            Repeater(repeatInterval = 10000L, shouldRepeat = repeatCount < repeats)
        }

    @Test
    fun test_custom_repeat_scheduler() {
        val testScheduler = TestScheduler()
        val repeats = 5
        val testObservable: TestObservableObserver<Long> =
            getRepeatObservable(repeats).flatMap {
                val interval = if (it.shouldRepeat) {
                    it.repeatInterval
                } else 0L
                singleTimer(interval, testScheduler)
            }.repeatUntil { interval -> interval == 0L }
                .subscribeOn(testScheduler)
                .test()

        testScheduler.timer.advanceBy(50000)
        assertEquals(repeats, testObservable.values.size)
    }

    private data class Repeater(val repeatInterval: Long, val shouldRepeat: Boolean)
```

### DIFF
--- a/reaktive-testing/src/commonMain/kotlin/com/badoo/reaktive/test/scheduler/TestScheduler.kt
+++ b/reaktive-testing/src/commonMain/kotlin/com/badoo/reaktive/test/scheduler/TestScheduler.kt
@@ -1,6 +1,7 @@
 package com.badoo.reaktive.test.scheduler
 
 import com.badoo.reaktive.scheduler.Scheduler
+import com.badoo.reaktive.scheduler.Scheduler.Executor
 import com.badoo.reaktive.utils.atomic.AtomicBoolean
 import com.badoo.reaktive.utils.atomic.AtomicLong
 import com.badoo.reaktive.utils.atomic.AtomicReference
@@ -12,16 +13,19 @@ class TestScheduler(
     private val isManualProcessing: Boolean = false
 ) : Scheduler {
 
-    val timer = Timer()
+    private val _timer = TimerImpl()
+    val timer: Timer = _timer
     private val _executors = AtomicReference<List<Executor>>(emptyList())
     val executors: List<Executor> get() = _executors.value
+    private val tasks = AtomicReference<List<Task>>(emptyList())
+    private var isProcessing = AtomicBoolean()
 
     init {
         freeze()
     }
 
-    override fun newExecutor(): Scheduler.Executor {
-        val executor = Executor(timer, isManualProcessing)
+    override fun newExecutor(): Executor {
+        val executor = ExecutorImpl()
         _executors.update { it + executor }
 
         return executor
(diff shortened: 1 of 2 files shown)


reaktive-testing/src/commonMain/kotlin/com/badoo/reaktive/test/scheduler/TestScheduler.kt
```diff
@@ -43,6 +45,13 @@ class TestScheduler(
         private val timeMillis = AtomicLong()
         private val listeners: AtomicReference<Set<() -> Unit>> = AtomicReference(emptySet())
         val millis: Long get() = timeMillis.value
+        private val minTimerStep = AtomicLong()
+
+        fun notifyDelay(startDelayMillis: Long) {
+            if (startDelayMillis > 0L) {
```
minaEweida, line 51: Save minimum delay to be used as step for advancing time

reaktive-testing/src/commonMain/kotlin/com/badoo/reaktive/test/scheduler/TestScheduler.kt
```diff
@@ -53,8 +62,16 @@ class TestScheduler(
         }
 
         fun advanceBy(millis: Long) {
-            timeMillis.addAndGet(millis)
-            listeners.value.forEach { it() }
+            var currentTimerStep = millis
+            while (currentTimerStep > 0) {
+                val advance = when (minTimerStep.value) {
+                    in 1 until currentTimerStep -> minTimerStep.value
```
minaEweida, line 68: if `minTimerStep` is between `1` till `millis` then keep advancing time with it, else just advance with the `millis`

reaktive-testing/src/commonMain/kotlin/com/badoo/reaktive/test/scheduler/TestScheduler.kt
```diff
@@ -116,6 +133,7 @@ class TestScheduler(
         }
 
         private fun addTask(startDelayMillis: Long, periodMillis: Long?, task: () -> Unit) {
+            timer.notifyDelay(startDelayMillis)
```
minaEweida, line 136: Updated `minTimerStep` with every task added

reaktive-testing/src/commonMain/kotlin/com/badoo/reaktive/test/scheduler/TestScheduler.kt
```diff
@@ -53,8 +62,18 @@ class TestScheduler(
         }
 
         fun advanceBy(millis: Long) {
-            timeMillis.addAndGet(millis)
-            listeners.value.forEach { it() }
+            var currentTimerStep = millis
+            while (currentTimerStep > 0) {
+                val advance =
```
minaEweida, line 67: if `minTimerStep` is between `1 until currentTimerStep` (which starts with `millis` as value) then keep advancing time with the `minTimerStep`, else just advance with the `currentTimerStep`

With every iteration `listeners.value.forEach { it() }` try to `processIfNeeded` to make sure that if we got a new task, it will be executed if needed

arkivanov:
Hello, thanks for the PR. This is a bug indeed. As of solution, I think we should try to fix it without adding any additional methods. Ideally it should just work, like in RxJava. I will think about it come back with my ideas.

arkivanov:
So I think the following implementation of the `TestScheduler` should work. @minaEweida please let me know what do you think.

This change breaks binary compatibility, but I don't think this is so important for the `reaktive-testing` module. I also removed public `Timer.addOnChangeListener` and `Timer.removeOnChangeListener` as they are not needed now. Overall these changes should not affect clients' builds in 99.9% of use cases.

@CherryPerry @mk-jones please let me know your opinions as well.

```kotlin
package com.badoo.reaktive.test.scheduler

import com.badoo.reaktive.scheduler.Scheduler
import com.badoo.reaktive.scheduler.Scheduler.Executor
import com.badoo.reaktive.utils.atomic.AtomicBoolean
import com.badoo.reaktive.utils.atomic.AtomicLong
import com.badoo.reaktive.utils.atomic.AtomicReference
import com.badoo.reaktive.utils.atomic.getAndUpdate
import com.badoo.reaktive.utils.atomic.update
import com.badoo.reaktive.utils.freeze

class TestScheduler(
    private val isManualProcessing: Boolean = false
) : Scheduler {

    private val _timer = TimerImpl()
    val timer: Timer = _timer
    private val _executors = AtomicReference<List<Executor>>(emptyList())
    val executors: List<Executor> get() = _executors.value
    private val tasks = AtomicReference<List<Task>>(emptyList())
    private var isProcessing = AtomicBoolean()

    init {
        freeze()
    }

    override fun newExecutor(): Executor {
        val executor = ExecutorImpl()
        _executors.update { it + executor }

        return executor
    }

    override fun destroy() {
        _executors
            .getAndUpdate { emptyList() }
            .forEach(Executor::dispose)
    }

    fun process() {
        if (isProcessing.compareAndSet(expectedValue = false, newValue = true)) {
            try {
                processActual()
            } finally {
                isProcessing.value = false
            }
        }
    }

    private fun processActual() {
        while (true) {
            val task = tasks.value.firstOrNull()?.takeIf { it.startMillis <= _timer.targetMillis } ?: break
            updateTasks {
                removeAt(0)
                if (task.periodMillis >= 0L) {
                    add(task.copy(startMillis = task.startMillis + task.periodMillis))
                    sort()
                }
            }

            _timer.millis = task.startMillis

            if (!task.executor.isDisposed) {
                task.task()
            }
        }

        _timer.millis = _timer.targetMillis
    }

    private fun processIfNeeded() {
        if (!isManualProcessing) {
            process()
        }
    }

    private inline fun updateTasks(block: MutableList<Task>.() -> Unit) {
        tasks.update {
            it.toMutableList().also(block)
        }
    }

    interface Timer {
        val millis: Long

        fun advanceBy(millis: Long)
    }

    private inner class TimerImpl : Timer {
        private val _millis = AtomicLong()
        override var millis: Long
            get() = _millis.value
            set(value) {
                _millis.value = value
            }

        private val _requestedMillis = AtomicLong()
        val targetMillis: Long get() = _requestedMillis.value

        override fun advanceBy(millis: Long) {
            require(millis >= 0L) { "Millis must not be negative" }

            _requestedMillis.addAndGet(millis)
            processIfNeeded()
        }
    }

    private inner class ExecutorImpl : Executor {
        private val _isDisposed = AtomicBoolean()
        override val isDisposed: Boolean get() = _isDisposed.value

        override fun submit(delayMillis: Long, task: () -> Unit) {
            addTask(startDelayMillis = delayMillis, task = task)
            processIfNeeded()
        }

        override fun submitRepeating(startDelayMillis: Long, periodMillis: Long, task: () -> Unit) {
            addTask(startDelayMillis = startDelayMillis, periodMillis = periodMillis, task = task)
            processIfNeeded()
        }

        private fun addTask(startDelayMillis: Long, periodMillis: Long = -1L, task: () -> Unit) {
            updateTasks {
                add(
                    Task(
                        startMillis = timer.millis + startDelayMillis,
                        periodMillis = periodMillis,
                        executor = this@ExecutorImpl,
                        task = task
                    )
                )
                sort()
            }
        }

        override fun cancel() {
            updateTasks {
                removeAll { it.executor === this@ExecutorImpl }
            }
        }

        override fun dispose() {
            _isDisposed.value = true
            cancel()
        }
    }

    private data class Task(
        val startMillis: Long,
        val periodMillis: Long,
        val executor: Executor,
        val task: () -> Unit
    ) : Comparable<Task> {
        private val sequenceNumber = sequencer.addAndGet(1L)

        override fun compareTo(other: Task): Int =
            if (this === other) {
                0
            } else {
                startMillis
                    .compareTo(other.startMillis)
                    .takeUnless { it == 0 }
                    ?: sequenceNumber.compareTo(other.sequenceNumber)
            }

        private companion object {
            private val sequencer = AtomicLong()
        }
    }
}
```

API change diff:
```
  @@ -172,24 +172,11 @@
   	public final fun process ()V
   }
   
  -public final class com/badoo/reaktive/test/scheduler/TestScheduler$Executor : com/badoo/reaktive/scheduler/Scheduler$Executor {
  -	public fun <init> (Lcom/badoo/reaktive/test/scheduler/TestScheduler$Timer;Z)V
  -	public fun cancel ()V
  -	public fun dispose ()V
  -	public fun isDisposed ()Z
  -	public final fun process ()V
  -	public fun submit (JLkotlin/jvm/functions/Function0;)V
  -	public fun submitRepeating (JJLkotlin/jvm/functions/Function0;)V
  +public abstract interface class com/badoo/reaktive/test/scheduler/TestScheduler$Timer {
  +	public abstract fun advanceBy (J)V
  +	public abstract fun getMillis ()J
   }
   
  -public final class com/badoo/reaktive/test/scheduler/TestScheduler$Timer {
  -	public fun <init> ()V
  -	public final fun addOnChangeListener (Lkotlin/jvm/functions/Function0;)V
  -	public final fun advanceBy (J)V
  -	public final fun getMillis ()J
  -	public final fun removeOnChangeListener (Lkotlin/jvm/functions/Function0;)V
  -}
  -
   public final class com/badoo/reaktive/test/scheduler/TestSchedulerExtKt {
   	public static final fun assertAllExecutorsDisposed (Lcom/badoo/reaktive/test/scheduler/TestScheduler;)V
   }
```

minaEweida:
Much better than my solution and more efficient since advancing will be exactly with the time needed while my fix might get stuck forever with the smallest step till a bigger delay comes in. May I suggest since this fix will break binary compatibility, might as well, change `timer` to private and just expose `advanceTime` in `TestScheduler` to make it simpler. I mean it's easier to just say `testScheduler.advanceTime` than `testScheduler.timer.advanceTime`

arkivanov:
@minaEweida I'm afraid we can't remove `Timer` because it will break compilations in 100% of cases. But we can add additional methods later in a separate PR. We will think about it. Thanks for the proposal.

So are you up to proceed fixing the issue? If so then please do the following:

- Revert your changes
- Apply and commit the suggested code of the `TestScheduler`
- Update your branch with the fresh master
- Run `:apiDump` task and commit changes

After that we will be able check the exact diff and will consider the binary compatibility concern.

Thanks for your contribution!

minaEweida:
Thanks for your help, I have rebased and ran `apiDump` to update with API changes. Please let me know if I missed anything

arkivanov:
> Thanks for your help, I have rebased and ran `apiDump` to update with API changes. Please let me know if I missed anything

Thanks! It will take a while to review the changes.

PR #473, opened 2020-05-31
Gradle 6 without parallel build

CherryPerry:
No more parallelisation, `buildSrc` is not available anymore in `settings.gradle`.


arkivanov:
> No more parallelisation, `buildSrc` is not available anymore in `settings.gradle`.

image

arkivanov:
How we will add Windows target later?

CherryPerry:
Same way as macOS.
Parallelisation will still exist, but JVM will be built on each OS.

No more _effective_ parallel builds.

PR #476, opened 2020-06-05
CI fixes

CherryPerry:
For some reason child java processes are stuck on GitHub CI (detekt, kotlin compiler).
I already encountered same issue, but in that case I created more than 10 of them.


CherryPerry:
@arkivanov Let's keep it, I will tweak it later. At least you are not blocked anymore.

PR #477, opened 2020-06-05
Avoid unnecessary freezing in xxxByEmmiter builders

### LINKED ISSUE #472: Freezing issue
I'm trying to track down a freezing issue on iOS when using reaktive streams, and pared down the issue with a new sample project.

It seems the call to mutate() below is causing ShimTester to be frozen, but I thought using threadLocal would prevent that. Is my current usage correct? What's the correct way to call mutate() in the example below?

KN code:

```

class ShimTester {

    init {
        expectEnsureNeverFrozenShim()
    }

    fun mutate() { }

    fun testApiWrapper(): ObservableWrapper<Unit> {
        return testApi()
                .subscribeOn(mainScheduler)
                .observeOn(mainScheduler)
                .threadLocal()
                .wrap()
    }

    fun testApi(): Observable<Unit> {
        return observableFromFunction {
            mutate()

            Unit
        }
    }
}
```

On iOS, I'm using:

```
        ShimTester().testApiWrapper().subscribe(
            isThreadLocal: true,
            onSubscribe: nil,
            onError: nil,
            onComplete: nil) { unit in
                print("Done")
        }
```

This causes an exception:

```
Uncaught exception: kotlin.native.concurrent.FreezingException: 
 freezing of [com.badoo.reaktive.scheduler.MainScheduler.ExecutorImpl.Operation@1e25bc8] has failed,
 first blocker is ...ShimTester@1daa928kotlin.native.concurrent.FreezingException:
 freezing of [com.badoo.reaktive.scheduler.MainScheduler.ExecutorImpl.Operation@1e25bc8] has failed,
 first blocker is ..ShimTester@1daa928
```

I tried moving threadLocal further upstream, before mutate(), but hit the same freeze exception:

```
class ShimTester {

    ...

    fun testApi(): Observable<Unit> {
        return observableOf(5)
                .subscribeOn(mainScheduler)
                .observeOn(mainScheduler)
                .threadLocal()
                .map {
                    mutate()
                    Unit
                }
    }
}

```

arkivanov:
Fixes #472


reaktive/src/nativeCommonMain/kotlin/com/badoo/reaktive/single/SingleByEmitter.kt
```diff
@@ -0,0 +1,47 @@
...
+
+import com.badoo.reaktive.base.tryCatch
+import com.badoo.reaktive.disposable.Disposable
+import com.badoo.reaktive.disposable.DisposableWrapper
+
+// Separate implementation prevents unnecessary freezing: https://github.com/badoo/Reaktive/issues/472
+actual inline fun <T> single(crossinline onSubscribe: (emitter: SingleEmitter<T>) -> Unit): Single<T> =
+    singleUnsafe { observer ->
+        val disposableWrapper = DisposableWrapper()
+        observer.onSubscribe(disposableWrapper)
```
CherryPerry, line 11: The only difference, that you extracted `DisposableWrapper` from `SingleEmitter`?
Can't we have this implementation for all platforms then?

arkivanov: Yes, the JVM version allocates less objects and so more optimal. I made this optimization when I was fighting with benchmarks.

arkivanov: @CherryPerry Whay do you think?

PR #482, opened 2020-06-07
Bump version to 1.1.14

arkivanov:


CherryPerry:
[1.3.72](https://github.com/JetBrains/kotlin/blob/master/ChangeLog.md#1372) is released, want to update before ours?

arkivanov:
> [1.3.72](https://github.com/JetBrains/kotlin/blob/master/ChangeLog.md#1372) is released, want to update before ours?

Filed a separate PR: #484

Issue #485, opened 2020-06-16
NodeJs: TypeError: window.setTimeout is not a function


minaEweida:
I recently started supporting Js on a library that uses reaktive and I had to publish it as an npm package and part of this process involved exporting all dependencies including Reaktive.
I was able to locally publish everything including Reaktive and initialize my library on a sample nodejs app but I got an error on my sample app:

`ReferenceError: window is not defined`

Because of this line (This is the javascript code that was publish not kotlin)
`var element = window.setTimeout(task, delayMillis.toInt());`

Which is actually in this function in `MainScheduler.kt` 
```
        override fun submit(delayMillis: Long, task: () -> Unit) {
            timeoutIds += window.setTimeout(task, delayMillis.toInt())
        }
```
I worked around it by doing this

```
global.window = {document: {createElementNS: () => {return {}} }};
```

But I got another error and I got stuck with it
```
TypeError: window.setTimeout is not a function
```

I am not nodejs expert so I wanted to ask, should the implementation work fine and the issue is from my side or the line below should be different in case of nodeJs not browser. Because I am assuming this should work fine on a browser? I would also assume `window.setInterval` should break for me as well.

```
timeoutIds += window.setTimeout(task, delayMillis.toInt())
```

arkivanov:
@minaEweida Thanks for reporting, this looks like a bug.

arkivanov:
@lukaville Would you like to take this?

minaEweida:
@arkivanov I did a fix that worked for me locally, I can share later today if it's not going to be fixed soon. Will send for review and then cleanup.

arkivanov:
@minaEweida Thanks, PR is very welcome!

arkivanov:
Closed by #486

PR #486, opened 2020-06-17
Fixing timer usage js

minaEweida:
Fix for https://github.com/badoo/Reaktive/issues/485

- Added external functions `setTimeout`, `setInterval`, `clearTimeout`, and `clearInterval`
- Created a Timers object that will handle calls
- If window is defined then use the window functions else use the external functions

There are 2 cases using where `window` will cause crash:
1- ReferenceError: window is not defined
2- TypeError: window.setTimeout is not a function (or setInterval or clearTimeout or clearInterval)

So for instance if someone works around window is not defined by doing the following:
`global.window = {document: {createElementNS: () => {return {}} }};`

If we still don't catch this error, MainScheduler will break.

So this approach will ensure that if window fails, we will fallback to external functions which should map to nodeJs timer functions


reaktive/src/jsMain/kotlin/com/badoo/reaktive/scheduler/MainScheduler.kt
```diff
@@ -19,30 +19,30 @@ internal class MainScheduler : Scheduler {
 
         private var _isDisposed = false
 
-        private val timeoutIds = mutableListOf<Int>()
-        private val intervalIds = mutableListOf<Int>()
+        private val timeoutIds = mutableListOf<Long>()
```
minaEweida, line 22: Not sure about changing to Long, but it made it easier to differentiate between 
`fun clearTimeout(interval: Long)` in `Timers` object and the external functions.

utils/src/jsMain/kotlin/com/badoo/reaktive/utils/Timers.kt
```diff
@@ -0,0 +1,42 @@
+package com.badoo.reaktive.utils
+
+import kotlin.browser.window
+
+external fun setTimeout(handler: dynamic, timeout: Int = definedExternally, vararg arguments: Any?): Int
```
arkivanov, line 5: Please make them `internal`

arkivanov: Why do we need `definedExternally` here?

utils/src/jsMain/kotlin/com/badoo/reaktive/utils/Timers.kt
```diff
@@ -0,0 +1,42 @@
...
+external fun setInterval(handler: dynamic, timeout: Int = definedExternally, vararg arguments: Any?): Int
+external fun clearTimeout(handle: Int = definedExternally)
+external fun clearInterval(handle: Int = definedExternally)
+
+object Timers {
+    fun setTimeout(delayMillis: Long, task: () -> Unit, vararg arguments: Any?): Int {
+        return try {
+            window.setTimeout(task, delayMillis.toInt(), arguments)
+        } catch (ex: dynamic) {
+            setTimeout(task, delayMillis.toInt(), arguments)
```
arkivanov, line 15: Avoiding clashing using different types is a bad workaround. Let's try the following:
1. Put `external` functions into `external` package
2. Import them into this file using named alias, e.g. `import xxx.external.setTimeout as setTimeoutExternal`
3. Remove `Timers` object and make these functions `internal` and top-level
3. Use `Int`

utils/src/jsMain/kotlin/com/badoo/reaktive/utils/Timers.kt
```diff
@@ -0,0 +1,42 @@
...
+external fun setTimeout(handler: dynamic, timeout: Int = definedExternally, vararg arguments: Any?): Int
+external fun setInterval(handler: dynamic, timeout: Int = definedExternally, vararg arguments: Any?): Int
+external fun clearTimeout(handle: Int = definedExternally)
+external fun clearInterval(handle: Int = definedExternally)
+
+object Timers {
+    fun setTimeout(delayMillis: Long, task: () -> Unit, vararg arguments: Any?): Int {
+        return try {
+            window.setTimeout(task, delayMillis.toInt(), arguments)
+        } catch (ex: dynamic) {
```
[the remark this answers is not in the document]
minaEweida: Actually that was my first approach in this commit https://github.com/badoo/Reaktive/pull/486/commits/c8c797ba7c7c31f28acaf6248582f6587ae230f5
But then if you define`window` as a workaround 
`global.window = {document: {createElementNS: () => {return {}} }};`
Or for some reason window is defined (not sure it's possible or not)
Then this fix will be useless because window will be defined so you will get the error 
`TypeError: window.setTimeout is not a function`

What do you think is better then? Should check if window is defined and try catch it just in case we get the other error?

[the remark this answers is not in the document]
arkivanov: Can you try `if (window == undefined) ... else ...` ?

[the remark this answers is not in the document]
lukaville: maybe we can use external setTimeout function in both browser and NodeJS environment?

utils/src/jsMain/kotlin/com/badoo/reaktive/utils/Timers.kt
minaEweida: just wanted to match the original contract. Will remove and test if it works, then I will remove it

utils/src/jsMain/kotlin/com/badoo/reaktive/utils/Timers.kt
minaEweida: @arkivanov I put those functions in `utils` module, so if I change them to internal I cannot reference from `reaktive` module. Would you like to not make them internal and keep in utils? OR should I move them to `reaktive` module and make them internal?

utils/src/jsMain/kotlin/com/badoo/reaktive/utils/Timers.kt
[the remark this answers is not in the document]
minaEweida: @lukaville I did not try since I am not testing against browser, but if you believe it will work for both, then I don't mind just directly calling it. What do you think?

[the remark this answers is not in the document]
minaEweida: @arkivanov in my case `window` was not `undefined` which cause the `setTimeout` is not a function error. So window can be undefined and still get the error in nodejs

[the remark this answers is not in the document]
arkivanov: Can you check somehow whether `window.setTimeout` is defined our not? `try-catch` is really a bad solution and we should try to find another way.

utils/src/jsMain/kotlin/com/badoo/reaktive/utils/Timers.kt
arkivanov: They are needed in the `reaktive` module so let's put them there (in `jsMain` source set) and make `internal`.

reaktive/src/jsMain/kotlin/com/badoo/reaktive/timer/Timers.kt
```diff
@@ -0,0 +1,21 @@
+package com.badoo.reaktive.timer
+
+import kotlin.browser.window
+import com.badoo.reaktive.timer.external.clearInterval as clearIntervalExternal
+import com.badoo.reaktive.timer.external.clearTimeout as clearTimeoutExternal
+import com.badoo.reaktive.timer.external.setInterval as setIntervalExternal
+import com.badoo.reaktive.timer.external.setTimeout as setTimeoutExternal
+
+private val isWindowDefined: Boolean = jsTypeOf(window) != "undefined"
```
minaEweida, line 9: This is verbose, could be better. However I tried to check if `setTimeout` is undefined but could not.

@lukaville Any idea how could it be done?

utils/src/jsMain/kotlin/com/badoo/reaktive/utils/Timers.kt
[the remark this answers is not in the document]
minaEweida: I could not do that, I just put window check for now. If calling setTimeout will work for browser and NodeJS then maybe just use it and ignore window

reaktive/src/jsMain/kotlin/com/badoo/reaktive/timer/Timers.kt
lukaville: I checked the docs and looks like the standard way to access global object in all environments is `globalThis`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis
https://www.codementor.io/@robertwozniak/to-clean-it-up-globalthis-is-here-x252w3rhh

It works in browsers and returns window:
```
> globalThis
Window {parent: Window, opener: null, top: Window, length: 0, frames: Window, …}
```

And it works in NodeJS as well:
```
> globalThis
Object [global] {
  global: [Circular],
  clearInterval: [Function: clearInterval],
  clearTimeout: [Function: clearTimeout],
  setInterval: [Function: setInterval],
  setTimeout: [Function: setTimeout] { [Symbol(util.promisify.custom)]: [Function] },
  queueMicrotask: [Function: queueMicrotask],
  clearImmediate: [Function: clearImmediate],
  setImmediate: [Function: setImmediate] {
    [Symbol(util.promisify.custom)]: [Function]
  }
}
```

minaEweida: Awesome, let me try that and get back to you

minaEweida:
```
external val globalThis: dynamic

internal fun setTimeout(task: () -> Unit, delayMillis: Int): Int =
    globalThis.setTimeout(task, delayMillis) as Int
```

using `globalThis.setTimeout` should return `Int` but if I try this it throws ClassCastException on nodejs
Screen Shot 2020-06-18 at 3 10 06 PM

Did I define it wrong here?
@lukaville

utils/src/jsMain/kotlin/com/badoo/reaktive/utils/Timers.kt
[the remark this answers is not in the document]
arkivanov: The `window` should not be defined in NodeJs. There some related explanations [here](https://stackoverflow.com/questions/45964178/referenceerror-window-is-not-defined-at-object-anonymous-node-js). I will try to find a way of how to check this in kotlin. The pure JS way is as follows: `if (typeof window !== 'undefined')`.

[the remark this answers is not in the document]
arkivanov: Maybe this could somehow work? https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.js/js-type-of.html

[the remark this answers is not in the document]
arkivanov: Also this may help: https://kotlinlang.org/docs/reference/js-interop.html#inline-javascript

reaktive/src/jsMain/kotlin/com/badoo/reaktive/timer/Timers.kt
arkivanov, line 9: So if this solutions works I'm fine with it. Can you write some tests? You could just write 4 tests each calling single method. Without any verification, just to make sure they do not crash. They will be run automatically for both browser and NodeJs.

minaEweida:
Updated PR with a smaller change

Using `globalThis`

- Given that globalThis will call the appropriate function for browser or nodejs it made sense to be used
- Changing list types for timeouts and intervals to `dynamic` because
`window.setTimeout` returns `number` however `setTimeout` for nodeJs returns an object
Which makes sense why I got ClasscastException https://github.com/badoo/Reaktive/pull/486#discussion_r442165500 when I assumed it returns `number`
- Created a global package for defining external `globalThis` object

reaktive/src/jsMain/kotlin/com/badoo/reaktive/global/external/ExternalGlobal.kt
```diff
@@ -0,0 +1,3 @@
+package com.badoo.reaktive.global.external
+
+external val globalThis: dynamic
```
arkivanov, line 3: Please make this internal.

reaktive/src/jsMain/kotlin/com/badoo/reaktive/timer/Timers.kt
```diff
@@ -0,0 +1,21 @@
...
+
+internal fun setTimeout(task: () -> Unit, delayMillis: Int): Int =
+    if (isWindowDefined) window.setTimeout(task, delayMillis) else setTimeoutExternal(task, delayMillis)
+
+internal fun setInterval(task: () -> Unit, delayMillis: Int): Int =
+    if (isWindowDefined) window.setInterval(task, delayMillis) else setIntervalExternal(task, delayMillis)
+
+internal fun clearTimeout(handler: Int) =
```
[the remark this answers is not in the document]
minaEweida: No need, removed the whole file

reaktive/src/jsMain/kotlin/com/badoo/reaktive/global/external/ExternalGlobal.kt
minaEweida: I cannot make internal, using the external keyboard to imply that it is not implemented by kotlin https://kotlinlang.org/docs/reference/js-interop.html#external-modifier

reaktive/src/jsMain/kotlin/com/badoo/reaktive/timer/Timers.kt
minaEweida: Removed the whole file, resolving

minaEweida: @lukaville Fixed it in the latest commit, resolving. Thanks for the suggestion

reaktive/src/jsMain/kotlin/com/badoo/reaktive/global/external/ExternalGlobal.kt
arkivanov: What about `internal external val globalThis: dynamic`? The thing is we don't need this be part of the public API of the library.

minaEweida: Done, amended. Thanks for your comments

PR #495, opened 2020-07-03
Add docs for interfaces with factory functions

### LINKED ISSUE #490: Add information about builder functions to corresponding interfaces
Currently it is not obvious how an interface can be instantiated. E.g. looking at the `Disposable` interface it is not easy to find the `Disposable {}` builder function. We should add this information to kdocs for all relevant interfaces.

arkivanov:
Closes #490

### DIFF
--- a/reaktive/src/commonMain/kotlin/com/badoo/reaktive/completable/Completable.kt
+++ b/reaktive/src/commonMain/kotlin/com/badoo/reaktive/completable/Completable.kt
@@ -4,6 +4,9 @@ import com.badoo.reaktive.base.Source
 
 /**
  * Represents a [Source] that either completes or produces an error.
+ *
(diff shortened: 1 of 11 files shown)


arkivanov:
> Something I noticed while going over these methods.
> For Observable factory methods, we have: `fun <T> Iterable<T>.asObservable()`
> We do have: `fun <T> observableOf(vararg values: T)`, but it's not quite the same.
> Do you think it is worth adding observableOfIterable, as otherwise people may miss it?

I was always trying to avoid duplicated API, e.g. we don't have `merge(Iterable<Observable>)`, but we do have `Iterable<Observable>.merge()` and `merge(vararg Observable)`. So we kinda consistent here.

But we could consider it as a separate PR, If you think it worth or there is a demand. What do you think?

PR #501, opened 2020-07-06
Fix empty collection should complete stream

### LINKED ISSUE #500: Collection<Observable<T>>.combineLatest will never complete if collection is empty
Hi, i faced a problem that `Observable.combineLatest` will not complete if collection of `Observable's` is empty.
This simple test case failed:
```kotlin
@Test
fun completed_when_collection_is_empty() {
    val observer = emptyList<Observable<Any>>()
        .combineLatest { it }
        .test()
    observer.assertNoValues()
    observer.assertComplete()
}
```

Also some of operators such as: `amb`, `merge`, `zip`, `withLatestFrom` also works incorrect. Whereas `concat` operators handle this case, so does [RxJava](https://github.com/ReactiveX/RxJava/blob/d209606e0eaaf87bb5493fd71a8413822f9992f2/src/main/java/io/reactivex/rxjava3/internal/operators/observable/ObservableCombineLatest.java#L70)
Furthermore there is different reciever type for collection: [Iterable](https://github.com/badoo/Reaktive/blob/6ae4e8caf147e784f21f16dcc49a6b8fe8ed6d92/reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/Concat.kt#L8) and [Collection](https://github.com/badoo/Reaktive/blob/6ae4e8caf147e784f21f16dcc49a6b8fe8ed6d92/reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/CombineLatest.kt#L13)

So i have some questions:
- Which type should be receiver: Iterable or Collection?
- Should i optimize operators with `size == 1` checks?
- Should i make single PR or per stream?

For example, what i want to do:
```kotlin
fun <T, R> Collection<Observable<T>>.combineLatest(mapper: (List<T>) -> R): Observable<R> =
    when (size) {
        0 -> observableOfEmpty()
        1 -> first().map { value -> mapper.invoke(List(1) { value }) }
        else -> internalCombineLatest(mapper)
    }

private fun <T, R> Collection<Observable<T>>.internalCombineLatest(mapper: (List<T>) -> R): Observable<R>
```

amihusb:
Fixes #500

### DIFF
--- a/reaktive/api/reaktive.api
+++ b/reaktive/api/reaktive.api
@@ -57,7 +57,7 @@ public final class com/badoo/reaktive/base/exceptions/TimeoutException : java/la
 }
 
 public final class com/badoo/reaktive/completable/AmbKt {
-	public static final fun amb (Ljava/util/Collection;)Lcom/badoo/reaktive/completable/Completable;
+	public static final fun amb (Ljava/lang/Iterable;)Lcom/badoo/reaktive/completable/Completable;
 	public static final fun amb ([Lcom/badoo/reaktive/completable/Completable;)Lcom/badoo/reaktive/completable/Completable;
 }
 
@@ -123,7 +123,7 @@ public final class com/badoo/reaktive/completable/CompletableWrapperKt {
 }
 
 public final class com/badoo/reaktive/completable/ConcatKt {
-	public static final fun concat (Ljava/util/Collection;)Lcom/badoo/reaktive/completable/Completable;
+	public static final fun concat (Ljava/lang/Iterable;)Lcom/badoo/reaktive/completable/Completable;
 	public static final fun concat ([Lcom/badoo/reaktive/completable/Completable;)Lcom/badoo/reaktive/completable/Completable;
 }
 
@@ -159,7 +159,7 @@ public final class com/badoo/reaktive/completable/DoOnBeforeKt {
 }
 
 public final class com/badoo/reaktive/completable/MergeKt {
-	public static final fun merge (Ljava/util/Collection;)Lcom/badoo/reaktive/completable/Completable;
+	public static final fun merge (Ljava/lang/Iterable;)Lcom/badoo/reaktive/completable/Completable;
 	public static final fun merge ([Lcom/badoo/reaktive/completable/Completable;)Lcom/badoo/reaktive/completable/Completable;
 }
 
@@ -293,7 +293,7 @@ public final class com/badoo/reaktive/disposable/scope/DisposableScopeExtKt {
 }
 
 public final class com/badoo/reaktive/maybe/AmbKt {
-	public static final fun amb (Ljava/util/Collection;)Lcom/badoo/reaktive/maybe/Maybe;
+	public static final fun amb (Ljava/lang/Iterable;)Lcom/badoo/reaktive/maybe/Maybe;
 	public static final fun amb ([Lcom/badoo/reaktive/maybe/Maybe;)Lcom/badoo/reaktive/maybe/Maybe;
 }
 
@@ -443,7 +443,7 @@ public final class com/badoo/reaktive/maybe/MaybeWrapperKt {
 }
 
 public final class com/badoo/reaktive/maybe/MergeKt {
-	public static final fun merge (Ljava/util/Collection;)Lcom/badoo/reaktive/observable/Observable;
+	public static final fun merge (Ljava/lang/Iterable;)Lcom/badoo/reaktive/observable/Observable;
 	public static final fun merge ([Lcom/badoo/reaktive/maybe/Maybe;)Lcom/badoo/reaktive/observable/Observable;
 }
 
@@ -547,7 +547,7 @@ public final class com/badoo/reaktive/maybe/ZipWithKt {
 }
 
 public final class com/badoo/reaktive/observable/AmbKt {
-	public static final fun amb (Ljava/util/Collection;)Lcom/badoo/reaktive/observable/Observable;
+	public static final fun amb (Ljava/lang/Iterable;)Lcom/badoo/reaktive/observable/Observable;
 	public static final fun amb ([Lcom/badoo/reaktive/observable/Observable;)Lcom/badoo/reaktive/observable/Observable;
 }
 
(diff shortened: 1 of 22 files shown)


amihusb:
I didn't find tests for `Observable.concat`, am i blind? If not i will write tests

arkivanov:
You are right, we don't have tests for `concat`, looks like we just missed them. But as you mentioned this operator is not affected by the "empty collection" bug. So I think tests for `concat` are out of scope of this PR. I have raised https://github.com/badoo/Reaktive/issues/502. Thanks!

amihusb:
There are something wrong with macOS build, can you explain what's wrong?

reaktive/api/reaktive.api
```diff
@@ -57,7 +57,7 @@ public final class com/badoo/reaktive/base/exceptions/TimeoutException : java/la
 }
 
 public final class com/badoo/reaktive/completable/AmbKt {
-	public static final fun amb (Ljava/lang/Iterable;)Lcom/badoo/reaktive/completable/Completable;
+	public static final fun amb (Ljava/util/Collection;)Lcom/badoo/reaktive/completable/Completable;
```
CherryPerry, line 60: @arkivanov 

Binary compatibility can be possibly broken. Not all Iterables are Collections.
But maybe it is fine to change it like this, not sure if anybody passes here pure custom Iterable implementation.

CherryPerry:
@amihusb 
```
> Task :reaktive:detekt
complexity - 20min debt
	LongMethod - 63/60 - [zip] at /Users/runner/work/Reaktive/Reaktive/reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/Zip.kt:15:38
```

reaktive/src/commonMain/kotlin/com/badoo/reaktive/completable/Amb.kt
```diff
@@ -3,8 +3,13 @@ package com.badoo.reaktive.completable
 import com.badoo.reaktive.base.CompositeDisposableObserver
 import com.badoo.reaktive.utils.atomic.AtomicBoolean
 
-fun Iterable<Completable>.amb(): Completable =
+fun Collection<Completable>.amb(): Completable =
```
arkivanov, line 6: Ok, I see we have different approaches in different such operators. E.g. in `concat` we are copying the `Iterable` into `List`. In `combineLatest` we have `Collection` and just iterating over it. In `amb` we have `Iterable` but we also need `isEmpty`.

So let's do the following: use `Iterable` everywhere and copy it `toList()` on susbcribe (like in `concat`). This will give us consistent API and all benefits of `Collection`. Also "consistency" with RxJava and probably safety (since collections can be potentially mutable under the hood).

arkivanov: I will wait for this to be applied and will continue reviewing the PR after that. Otherwise please let me know. Thanks!

amihusb: That's a good point, but with `receiver` and `sources` it's easy to make a mistake and for example to call `forEach` on `receiver`, which can work wrong only with mutable receiver.

In my opinion there are four solutions:
- inside "subscribe" wrap everything in `with(toList())` (i prefer this one)
- write test for every such case with mutable receiver
- write two functions: `fun Iterable.amb() = toList.amb()` and `private fun List.amb()`
- leave this to reviewers

arkivanov: Sorry for a delay with the response. I think the `concat` is already fine and there is no need to touch it. Affected operators iterate only once so just be careful there. Please keep in mind that collections should be converted lazily - inside `subscribe` block.

amihusb: Ok, done.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/completable/Merge.kt
```diff
@@ -33,13 +40,9 @@ fun Iterable<Completable>.merge(): Completable =
                 }
             )
         }
-
-        if (activeSourceCount.addAndGet(-1) == 0) {
```
arkivanov, line 37: This change will brake the operator in case when one of the upstreams will complete while subscribing to others. The following test demonstrates the issue, could you please add it as well? The `toList` conversion in this particular operator is therefore unnecessary.

```kotlin
    @Test
    fun does_not_complete_WHEN_one_upstream_completed_while_subscribing_to_others() {
        val observer =
            merge(
                    observer.onSubscribe(Disposable())
                    observer.onComplete()
                },
                TestCompletable()
            ).test()

        observer.assertNotComplete()
    }
```

reaktive/src/commonMain/kotlin/com/badoo/reaktive/maybe/Merge.kt
```diff
@@ -41,13 +48,9 @@ fun <T> Iterable<Maybe<T>>.merge(): Observable<T> =
                 }
             )
         }
-
-        if (activeSourceCount.addAndGet(-1) == 0) {
```
arkivanov, line 45: Same here. Could please adopt and add the provided test above for this operator?

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/Zip.kt
```diff
@@ -110,7 +117,7 @@ private sealed class ZipEvent<out T> {
 
 fun <T, R> zip(vararg sources: Observable<T>, mapper: (List<T>) -> R): Observable<R> =
     sources
-        .toList()
+        .asList()
```
arkivanov, line 120: :+1:

reaktive/src/commonMain/kotlin/com/badoo/reaktive/completable/Merge.kt
amihusb: Yes, i will add this test. My bad, didn't test this operator before fix.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/maybe/Merge.kt
amihusb: Done.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/completable/Merge.kt
amihusb: I think that `completableOfEmpty` is better due less code, check for `isDisposed` inside `completableOfEmpty` can't affect this, right?

arkivanov: We are trying to avoid unrelated production code in tests. Because we wan't to test exactly what is under test, and not other (unrelated) code. If `observableOfEmpty` at some point will be broken then this particular test will be red, which ideally should not happen.

arkivanov: Would you mind to use `completableUnsafe` instead? After that I think we are done.

amihusb: Ok.

PR #508, opened 2020-07-21
Optimize native BufferedExecutor

arkivanov:
This PR improves performance 100x-500x of the following code in K/N:
```kotlin
    observable<Int> { emitter ->
        repeat(10000, emitter::onNext)
        emitter.onComplete()
    }
        .observeOn(mainScheduler)
        .subscribe(onComplete = {})
```


reaktive/src/nativeCommonMain/kotlin/com/badoo/reaktive/utils/RefCounter.kt
```diff
@@ -0,0 +1,21 @@
...
+    private val destroy: () -> Unit
+) {
+
+    private val count = AtomicInt(1)
+
+    fun retain(): Boolean =
+        count.updateAndGet { if (it > 0) it + 1 else 0 } > 0
+
+    fun release() {
+        val newCount = count.updateAndGet { if (it > 0) it - 1 else throw IllegalStateException("RefCounter is already destroyed") }
```
CherryPerry, line 16:
```kotlin
count.updateAndGet { 
    check(it <= 0) { "RefCounter is already destroyed" }
    it - 1
}
```

reaktive/src/nativeCommonMain/kotlin/com/badoo/reaktive/scheduler/BufferedExecutor.kt
```diff
@@ -1,50 +1,60 @@
...
+    }
+
     actual fun submit(value: T) {
-        val oldState =
-            state.getAndUpdate {
-                it.copy(
-                    queue = it.queue + value,
-                    isDraining = true
-                )
+        refCounter.use {
```
mk-jones, line 32: Unlike in the jvm BufferedExecutor it is not possible to continue using a disposed executor and will throw an exception here.

arkivanov: The `RefCounter` throws an exception only if `release()` is called when the `RefCounter` if destroyed (ref count reached 0). If the `retain()` call returned `false` then the block won't be executed and the `release()` won't be called, and so wan't crash.

mk-jones: Right, my mistake 👍 . Mixed the two up.

In that case I think both shouldn't accept input when disposed.

reaktive/src/jvmJsCommonMain/kotlin/com/badoo/reaktive/scheduler/BufferedExecutor.kt
```diff
@@ -1,18 +1,26 @@
...
 
+    override var isDisposed: Boolean = false
+        private set
+
+    override fun dispose() {
+        isDisposed = true
+    }
+
     actual fun submit(value: T) {
         synchronized(monitor) {
```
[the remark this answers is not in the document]
arkivanov: The purpose of the `RefCounter` here is to protected the `Lock` instance from being used when it is already destroyed. Since the `Lock` is a resource in K/N, it requires explicit destroy. It is illegal to use the native `Lock` if its memory is reclaimed. There is no intention to protect the executor itself, it is just delegates calls to the consumer (where the protection should present).

Perhaps I should add a comment for `RefCounter` use case here?

PR #511, opened 2020-07-27
Check publication on CI

arkivanov:


.github/workflows/release.yml
```diff
@@ -31,5 +31,35 @@ jobs:
...
+      - name: Install Java
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+      - name: Install dependencies
+        run: sudo apt-get install nodejs chromium-browser
+      - name: Validate Gradle Wrapper
+        uses: gradle/wrapper-validation-action@v1
+      - name: Check JVM, Android, JS, Linux
+        run: ./gradlew :tools:check-publication:build -Ptarget=all_linux_hosted -Pbintray_user=${{ secrets.BINTRAY_USER }} -Pbintray_key=${{ secrets.BINTRAY_KEY }} -Dcheck_publication=true
```
CherryPerry, line 52: Can we use project property instead of system property?
You can use `findProperty` or `hasProperty` method.

settings.gradle.kts
```diff
@@ -23,3 +23,7 @@ include(":sample-js-browser-app")
 include(":sample-linuxx64-app")
 include(":sample-ios-app")
 include(":sample-macos-app")
+
+if (System.getProperty("check_publication")?.toBoolean() == true) {
```
CherryPerry, line 27: You can compare with string, it is easier.

includedBuild/gradleConfiguration/src/main/kotlin/com/badoo/reaktive/compatibility/BinaryCompatibilityConfigurationPlugin.kt
```diff
@@ -13,15 +13,16 @@ class BinaryCompatibilityConfigurationPlugin : Plugin<Project> {
...
                         "benchmarks",
                         "jmh",
                         "sample-mpp-module",
                         "sample-android-app",
                         "sample-js-browser-app",
                         "sample-linuxx64-app",
                         "sample-ios-app",
-                        "sample-macos-app"
+                        "sample-macos-app",
+                        if (System.getProperty("check_publication")?.toBoolean() == true) "check-publication" else null
```
CherryPerry, line 25: `"check-publication".takeIf { target.hasProperty("check_publication") }`

.github/workflows/release.yml
```diff
@@ -31,5 +31,35 @@ jobs:
...
+  macos-host-check:
+    name: Check on macOS
+    runs-on: macOS-latest
+    needs: macos-host-publish
```
CherryPerry, line 56: Needs both macos and linux, because linux publishes metadata, that can be not available yet.

CherryPerry:
Sorry, I pressed "approved". Do not merge.

.github/workflows/release.yml
arkivanov: I tried, these functions are not resolved in settings.gradle, for some reason.

CherryPerry: Try the following please:
```kotlin
val check_publication: String? by settings
if (check_publication == "true") {
...
}
```

CherryPerry: Issue: https://github.com/gradle/kotlin-dsl-samples/issues/635

arkivanov:
Fixed all comments

.github/workflows/release.yml
```diff
@@ -31,5 +31,35 @@ jobs:
...
+    needs: [linux-host-publish, macos-host-publish]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Install Java
+        uses: actions/setup-java@v1
+        with:
+          java-version: 1.8
+      - name: Check iOS
+        run: ./gradlew :tools:check-publication:build -Ptarget=all_macos_hosted -Pbintray_user=${{ secrets.BINTRAY_USER }} -Pbintray_key=${{ secrets.BINTRAY_KEY }} -Dcheck_publication=true
```
CherryPerry, line 65: -Pcheck_publication=true

.github/workflows/release.yml
CherryPerry, line 52: -Pcheck_publication=true

arkivanov: Fixed both. Actually just `-Pcheck_publication` is enough in this case.

Issue #411, opened 2020-02-11
Add Observable.window operator


arkivanov:

amihusb:
I want to implement this.
```kotlin
fun <T> Observable<T>.window(
    count: Long,
    skip: Long = count,
    bufferSize: Int = Int.MAX_VALUE
): Observable<Observable<T>> = TODO()

fun <T> Observable<T>.window(
    spanMillis: Long,
    scheduler: Scheduler,
    skipMillis: Long = spanMillis,
    maxSize: Long = Long.MAX_VALUE,
    restartOnMaxSize: Boolean = false,
    bufferSize: Int = Int.MAX_VALUE
): Observable<Observable<T>> = TODO()

fun <T> Observable<T>.window(
    boundary: Observable<*>,
    bufferSize: Int = Int.MAX_VALUE
): Observable<Observable<T>> = TODO()

fun <T, U> Observable<T>.window(
    openingIndicator: Observable<U>,
    closingIndicator: (U) -> Maybe<*>,
    bufferSize: Int = Int.MAX_VALUE
): Observable<Observable<T>> = TODO()

fun <T> Observable<T>.window(
    boundarySupplier: () -> Maybe<*>,
    bufferSize: Int = Int.MAX_VALUE
): Observable<Observable<T>> = TODO()
```

I will start with the first variant.

- Are there all overloads of window operator?
- Do you want to change something in signature or names?
- Can we reduce the number of implementations? How? I didn't find any way
- Should we rename the second overload to `windowTimed`?

arkivanov:
@amihusb I need to think about the API a bit, I will try to respond by 31/07.

arkivanov:
Hey @amihusb, I have found a draft stash of the `window` operator on my machine. I could try to finish it tomorrow, at least one of the overloads. Would it work for you?

amihusb:
@arkivanov yes. Which overload will you do?

arkivanov:
@amihusb I still need to think about the API, because `window` is not a straightforward operator. But when I was working on the draft I had the following in mind:

```kotlin
fun <T> Observable<T>.window(count: Int, skip: Int = count): Observable<Observable<T>> = TODO()

fun <T> Observable<T>.window(boundary: Single<WindowBoundary>): Observable<Observable<T>> = TODO()

fun <T> Observable<T>.window(boundary: Observable<WindowBoundary>): Observable<Observable<T>> = TODO()

class WindowBoundary(
    val closingSignal: Completable,
[...]

PR #516, opened 2020-08-08
Added Observable.window(boundaries) operator

arkivanov:
As part of #411 

Operators implemented:

- http://reactivex.io/RxJava/javadoc/io/reactivex/Observable.html#window-io.reactivex.ObservableSource-
- http://reactivex.io/RxJava/javadoc/io/reactivex/Observable.html#window-long-java.util.concurrent.TimeUnit-long-boolean-

Some tricky behaviours:

- The `upstream` _should not_ be disposed if the `downstream` is disposed but the last window is still subscribed, but the `boundaries` stream _should_ be disposed. If then the last window is disposed, the `upstream` should be disposed as well.
- If either `upstream` or `boundaries` stream is completed then everything should be cancelled and completed as well
- `UnicastSubject` is used as window, so it can be subscribed only once but all values are accumulating until the window is susbcribed

### DIFF
--- a/reaktive/api/reaktive.api
+++ b/reaktive/api/reaktive.api
@@ -946,6 +946,13 @@ public final class com/badoo/reaktive/observable/VariousKt {
 	public static final fun toObservableOfError (Ljava/lang/Throwable;)Lcom/badoo/reaktive/observable/Observable;
 }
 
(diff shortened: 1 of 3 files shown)


arkivanov:
@CherryPerry @mk-jones Please review the PR when you have time :-)

PR #518, opened 2020-08-10
Add Completable.testAwait() testing function

arkivanov:


reaktive-testing/src/commonMain/kotlin/com/badoo/reaktive/test/completable/TestAwait.kt
```diff
@@ -0,0 +1,32 @@
+package com.badoo.reaktive.test.completable
+
+import com.badoo.reaktive.completable.Completable
+import com.badoo.reaktive.completable.blockingAwait
+
+/**
+ * This method is used when you need to wait for an asynchronous operation to finish
```
arkivanov, line 7: One of our users (@omainegra) had this use case. @CherryPerry What do you think about this?

reaktive-testing/src/commonMain/kotlin/com/badoo/reaktive/test/completable/TestAwait.kt
```diff
@@ -0,0 +1,32 @@
...
+ * and you can't use or don't want to use the [TestScheduler][com.badoo.reaktive.test.scheduler.TestScheduler].
+ *
+ * It is JavaScript friendly, it returns `Promise` which is when returned from a test method
```
mk-jones, line 10: "...it returns `Promise` which, when returned from a test method, causes..."

reaktive-testing/src/commonMain/kotlin/com/badoo/reaktive/test/completable/TestAwait.kt
mk-jones: Looks good to me. However I'd say it looks like the TestScheduler should be preferred.

arkivanov: Absolutely, in 99% of the cases. But there are some tricky cases when this may be needed.

reaktive-testing/src/commonMain/kotlin/com/badoo/reaktive/test/completable/TestAwait.kt
```diff
@@ -0,0 +1,32 @@
...
+ *         singleFromFunction { a + b }
+ *             .subscribeOn(computationScheduler)
+ * }
+ *
+ * class CalculatorTest {
+ *     @Test
+ *     fun sum_2_3_returns_5() =
+ *         Calculator()
+ *             .sum(2, 3)
+ *             .doOnBeforeSuccess { assertEquals(6, it) }
```
CherryPerry, line 26: This part looks bad. Can't we do something like in code below?

```kotlin
Calculator()
    .sum(2, 3)
    .test()
    .await() // extension on test subscriber, not on Completable or whatever else
    .assertValue(6)
```

Or I am missing the point?

arkivanov: Good point. How about the following?

`expect fun <T> Single<T>.testAwait(onSuccess: (T) -> Unit = {})`

```kotlin
class MyLogicTest {

    @Test
    fun foo() =
        MyLogic()
            .sum(1, 2)
            .testAwait { assertEquals(3, it) }
}
```

reaktive-testing/src/commonMain/kotlin/com/badoo/reaktive/test/completable/TestAwait.kt
omainegra: Yes, I usually prefer `TestScheduler`, but I was writing an Integration Test so I needed a real scheduler, and I couldn't block the test in JS

arkivanov: @omainegra I also noticed that tests are executed on the main thread for e.g. iOS. So the following code will cause dead lock:
```kotlin
@Test
fun foo() =
    singleOf(1)
        .observeOn(mainThread)
        .testAwait()
```
In Android this code will fail since `Looper` is not instrumented.

So this is omething to keep in mind. The `mainScheduler` scheduler has to be overriden anyway.

reaktive-testing/src/commonMain/kotlin/com/badoo/reaktive/test/completable/TestAwait.kt
CherryPerry: But is not it impossible to check errors in your case? Let's return from `testAwait()` test subscriber?

arkivanov: We must return `Promise` for JS. This special case is recognized by the framework and it will wait untill the `Promise` resolves or fails.

reaktive-testing/src/commonMain/kotlin/com/badoo/reaktive/test/completable/TestAwait.kt
omainegra: Good to know, thank you

reaktive-testing/src/commonMain/kotlin/com/badoo/reaktive/test/completable/TestAwait.kt
arkivanov: If you need to verify error. Then the error should be a positive case. Like:

```kotlin
   @Test
    fun foo() =
        MyLogic()
            .div(5, 0)
            .onErrorReturn { it }
            .testAwait { assertTrue(it is ArithmeticException) }
```

arkivanov: Or we could add additional argument:
`expect fun <T> Single<T>.testAwait(assertSuccess: (T) -> Unit = {}, assertError: ((Throwable) -> Unit)? = null)`

arkivanov:
```kotlin
    @Test
    fun foo() =
        MyLogic()
            .div(5, 0)
            .testAwait(assertError = { assertTrue(it is ArithmeticException) }, assertSuccess = { fail() })
```

omainegra: Looks sweet, I like the `assertError` is optional since we never use `onError`, early on the pipeline we convert it to an ADT, like in `Single<Either<ErrorADT, T>`

arkivanov: Yeah. The idea of the optional `assertError` callback is: if it is provided then it must fail to make the test failed. If `assertError` succeeds then the error is considered as expected.

reaktive-testing/src/commonMain/kotlin/com/badoo/reaktive/test/completable/TestAwait.kt
arkivanov: Would you mind to recheck spelling one more time? And the code as well.

arkivanov:
@mk-jones Could please check the fixes?

PR #521, opened 2020-08-16
Update Kotlin to 1.4.0

arkivanov:


CherryPerry:
Can you check also settings.gradle.kts? I have feeling, that it also has kotlin snapshot repo link.

PR #531, opened 2020-08-27
Automate nmtc

arkivanov:
I think we'll need to publish mt coroutines interop multiple times, so would be nice to automate. The idea is to create a tag in the `mt-cotourines` branch every time after a release and have this workflow executed. Not sure about the yaml filter.

### DIFF
--- a/.github/workflows/release-nmtc.yml
+++ b/.github/workflows/release-nmtc.yml
@@ -1,11 +1,7 @@
 name: Release nmtc
 
 on:
-  push:
-    branches:
-      - 'mt-coroutines'
-    tags:
-      - '*'
+  workflow_dispatch
 
 jobs:
   linux-host-publish:
--- a/.github/workflows/release-nmtc.yml
+++ b/.github/workflows/release-nmtc.yml
@@ -1,0 +1,39 @@
+name: Release nmtc
+
+on:
+  push:
+    branches:
+      - 'mt-coroutines'
+    tags:
+      - '*'
+
+jobs:
+  linux-host-publish:
+    name: Release on Ubuntu
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Install Java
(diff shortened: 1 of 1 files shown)


CherryPerry (approved):
Should work I think.

.github/workflows/release-nmtc.yml
```diff
@@ -0,0 +1,39 @@
+name: Release nmtc
+
+on:
```
CherryPerry, line 3: Let's create releases as it is intended.

CherryPerry (changes requested):

.github/workflows/release-nmtc.yml
CherryPerry: Let's use [workflow_dispatch](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_dispatch). [Blog](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/).

```yaml
on:
  workflow_dispatch
```

Special GUI on website should appear.

.github/workflows/release-nmtc.yml
```diff
@@ -0,0 +1,39 @@
...
+  push:
```
CherryPerry, line 4:
```
on:
  workflow_dispatch
```

https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/

CherryPerry (changes requested):

.github/workflows/release-nmtc.yml
CherryPerry: We do not need any parameters, but if it fails to appear/run, we can add some random one.

arkivanov: Awesome!!! Will do

CherryPerry (approved):

CherryPerry:
Screenshot 2020-08-27 at 22 07 48

PR #534, opened 2020-08-30
Added Observable.window(count, skip) operator

amihusb:
As part of #411

Feels like i don't cover all cases in tests. Too complicated operator.

`WindowSized` is the appropriate name?

I'm embarrassed for the long wait, relocated to another city this month.

### DIFF
--- a/reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/WindowSized.kt
+++ b/reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/WindowSized.kt
@@ -21,76 +21,95 @@ fun <T> Observable<T>.window(
     require(skip > 0) { "skip > 0 required but it was $skip" }
 
     return observable { emitter ->
-        val upstreamDisposableWrapper = DisposableWrapper()
-
-        val windows = SharedQueue<UnicastSubject<T>>()
         val activeWindowsCount = AtomicInt(1)
+        val upstreamObserver = UpstreamObserver(
+            count = count,
+            skip = skip,
+            activeWindowsCount = activeWindowsCount,
+            emitter = emitter
+        )
 
         emitter.setCancellable {
             if (activeWindowsCount.addAndGet(-1) == 0) {
-                upstreamDisposableWrapper.dispose()
+                upstreamObserver.dispose()
             }
         }
 
-        val onTerminate: () -> Unit = {
-            if (activeWindowsCount.addAndGet(-1) == 0 && emitter.isDisposed) {
-                upstreamDisposableWrapper.dispose()
-            }
+        subscribe(upstreamObserver)
+    }
+}
+
+private class UpstreamObserver<T>(
+    private val count: Long,
+    private val skip: Long,
+    private val activeWindowsCount: AtomicInt,
+    private val emitter: ObservableEmitter<Observable<T>>
+) : DisposableWrapper(), ObservableObserver<T> {
+
+    private val windows = SharedQueue<UnicastSubject<T>>()
+    private val skippedCount = AtomicLong()
+    private val tailWindowValuesCount = AtomicLong()
+
+    private val onWindowTerminate: () -> Unit = {
+        if (activeWindowsCount.addAndGet(-1) == 0 && emitter.isDisposed) {
+            dispose()
+        }
+    }
+
+    override fun onSubscribe(disposable: Disposable) {
+        set(disposable)
+    }
+
+    override fun onNext(value: T) {
+        val skipped = skippedCount.value
+        val windowWrapper: WindowWrapper<T>?
+
+        if (skipped == 0L) {
+            activeWindowsCount.addAndGet(1)
+            val window = UnicastSubject<T>(onTerminate = onWindowTerminate)
+            windowWrapper = WindowWrapper(window)
+            windows.offer(window)
+            emitter.onNext(windowWrapper)
+        } else {
+            windowWrapper = null
         }
 
-        subscribe(object : ObservableObserver<T> {
+        windows.forEach { it.onNext(value) }
 
-            private val skippedCount = AtomicLong()
-            private val tailWindowValuesCount = AtomicLong()
+        skippedCount.value = (skipped + 1) % skip
 
-            override fun onSubscribe(disposable: Disposable) {
-                upstreamDisposableWrapper.set(disposable)
-            }
+        if (tailWindowValuesCount.value + 1 == count) {
+            requireNotNull(windows.poll()).onComplete()
+            tailWindowValuesCount.addAndGet(1 - skip)
+        } else {
+            tailWindowValuesCount.addAndGet(1)
+        }
 
-            override fun onNext(value: T) {
-                val skipped = skippedCount.value
+        if (windowWrapper?.isSubscribed?.value == false) {
+            windowWrapper.window.onComplete()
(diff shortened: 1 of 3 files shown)


arkivanov:
Thanks for the PR, I will check when I have time. From the first sight, looks like you can avoid using the Serializer, since there is only one stream - the upstream. Serializer is needed when there are multiple concurrent data streams. The upstream is guaranteed to produce items sequentially.

amihusb:
Oh, yes, i can avoid serializer, thanks. Then i need an atomic counter of active windows.

I don't miss anything: upstream is guaranteed to produce items sequentially by library operators/factories, but we can create "unstable" upstream by `observableUnsafe { ...creating threads and call observer.onNext... }`, right?

Will update implementation today.

arkivanov:
Sequential streams are guaranteed by the Rx contract. Every stream is allowed to emit items on different threads but always sequentially. You can break the contract of course, but operators should not cover such cases.

amihusb:
@arkivanov thanks for the clarification. Updated PR.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/WindowSized.kt
```diff
@@ -0,0 +1,96 @@
...
+        val onTerminate: () -> Unit = {
+            if (activeWindowsCount.addAndGet(-1) == 0 && emitter.isDisposed) {
+                upstreamDisposableWrapper.dispose()
+            }
+        }
+
+        subscribe(object : ObservableObserver<T> {
+
+            private val skippedCount = AtomicLong()
+            private val tailWindowValuesCount = AtomicLong()
```
arkivanov, line 44: I think just one counter should be enough. Suppose this `counter` would count total amount of emitter items. Then the following math should be valid?
1. We should open a new window if `counter % skip == 0`
2. We should close an oldest window if:
```
val openIndex = counter - count + 1
(openIndex >= 0) && (openIndex % skip == 0)
```

Please correct me if I'm wrong.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/WindowSized.kt
```diff
@@ -0,0 +1,96 @@
...
+                skippedCount.value = (skipped + 1) % skip
+
+                if (tailWindowValuesCount.value + 1 == count) {
+                    requireNotNull(windows.poll()).onComplete()
+                    tailWindowValuesCount.addAndGet(1 - skip)
+                } else {
+                    tailWindowValuesCount.addAndGet(1)
+                }
+
+                if (window != null && windowSubscribed != null && windowSubscribed.compareAndSet(false, true)) {
```
arkivanov, line 78: No need to `compareAndSet`, just check `!windowSusbcribed.value`

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/WindowSized.kt
```diff
@@ -0,0 +1,96 @@
...
+            }
+        }
+
+        val onTerminate: () -> Unit = {
+            if (activeWindowsCount.addAndGet(-1) == 0 && emitter.isDisposed) {
+                upstreamDisposableWrapper.dispose()
+            }
+        }
+
+        subscribe(object : ObservableObserver<T> {
```
arkivanov, line 41: Let's extract this `ObservableObserver` to a separate `private class`

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/WindowSized.kt
```diff
@@ -0,0 +1,96 @@
...
+
+            override fun onSubscribe(disposable: Disposable) {
+                upstreamDisposableWrapper.set(disposable)
+            }
+
+            override fun onNext(value: T) {
+                val skipped = skippedCount.value
+
+                val window: UnicastSubject<T>?
+                val windowSubscribed: AtomicBoolean?
```
arkivanov, line 54: We could introduce a separate private wrapper class, same as in RxJava and as I did here: #535 
In this case you will have only variable to check for null. Also this method will be cleaner.

arkivanov (changes requested):

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/WindowSized.kt
amihusb: Nice catch! Fixed.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/WindowSized.kt
amihusb: Done.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/WindowSized.kt
amihusb: Done.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/WindowSized.kt
amihusb: This math looks correct. Nice!

I spent a lot of time to find a way to use one counter. I checked the [Reactor](https://github.com/reactor/reactor-core/blob/master/reactor-core/src/main/java/reactor/core/publisher/FluxWindow.java#L459) and the [RxJava](https://github.com/ReactiveX/RxJava/blob/1ff3cd62091db049b39a2eda0443240e75d6c321/src/main/java/io/reactivex/rxjava3/internal/operators/observable/ObservableWindow.java#L149) implementations and they both use two counters.

arkivanov: Sorry for long delay. Could you please provide an example of an overflow?

arkivanov: Got it. This is a very good catch, let's use two counters. Sorry for confusing.

arkivanov (changes requested):
Sorry, let's use two counters. Otherwise looks good!

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/WindowSized.kt
```diff
@@ -0,0 +1,110 @@
...
+private class UpstreamObserver<T>(
+    private val count: Long,
+    private val skip: Long,
+    private val activeWindowsCount: AtomicInt,
+    private val emitter: ObservableEmitter<Observable<T>>
+) : DisposableWrapper(), ObservableObserver<T> {
+    private val windows = SharedQueue<UnicastSubject<T>>()
+    private val counter = AtomicLong()
+    private val onWindowTerminate: () -> Unit = {
+        if (activeWindowsCount.addAndGet(-1) == 0 && emitter.isDisposed) {
```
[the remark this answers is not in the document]
amihusb: Yes, it is useless now, thanks. It were necessary before `activeWindowsCount`.

arkivanov (approved):

arkivanov:
@amihusb Thanks for your contribution!

PR #544, opened 2020-10-03
Remove "apt get install" from check publication

arkivanov:
The job fails without "apt get update", but I think we don't need these packages at all.


idyatlov:
W

CherryPerry (approved):

PR #553, opened 2020-10-28
add weak

### LINKED ISSUE #388: iOS sample code contains retain loop
The iOS sample code contains a retain loop. The [ViewController](https://github.com/badoo/Reaktive/blob/5ec83b4244e25783fc70b50b547212c12b40cefc/sample-ios-app/sample-ios-app/ViewController.swift) owns the KittenView and the KittenView owns the ViewController. Personally I am not very familiar with the used approach which is not entirely idiomatic for Swift/iOS so not entirely sure what the best fix would be.

AlexeyKorshun:
Fixes #388


arkivanov:
Thanks for the PR! The build failed with the following error:
```
2020-10-28T05:08:32.1784750Z /Users/runner/work/Reaktive/Reaktive/sample-ios-app/sample-ios-app/ViewController.swift:34:26: error: 'weak' must be a mutable variable, because it may change at runtime
2020-10-28T05:08:32.1785870Z ** BUILD FAILED **
2020-10-28T05:08:32.1789390Z 
2020-10-28T05:08:32.1789810Z         private weak let viewController: ViewController
2020-10-28T05:08:32.1844470Z 
2020-10-28T05:08:32.1844910Z                          ^
2020-10-28T05:08:32.1847240Z /Users/runner/work/Reaktive/Reaktive/sample-ios-app/sample-ios-app/ViewController.swift:34:26: error: 'weak' variable should have optional type 'ViewController?'
2020-10-28T05:08:32.1851620Z The following build commands failed:
2020-10-28T05:08:32.1853360Z 	CompileSwift normal x86_64 /Users/runner/work/Reaktive/Reaktive/sample-ios-app/sample-ios-app/ViewController.swift
2020-10-28T05:08:32.1854730Z 	CompileSwiftSources normal x86_64 com.apple.xcode.tools.swift.compiler
2020-10-28T05:08:32.1869840Z (2 failures)
2020-10-28T05:08:32.1886710Z         private weak let viewController: ViewController
2020-10-28T05:08:32.1887630Z                          ^
2020-10-28T05:08:32.1887870Z                                                        ?
```

Would you mind fixing it?

AlexeyKorshun:
@arkivanov yes it is my night fault.

AlexeyKorshun:
@arkivanov can you help with error on CI? I see this

```
** BUILD SUCCEEDED **


FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':sample-ios-app:build'.
> Process 'command 'xcodebuild'' finished with non-zero exit value 65

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 58m 36s
```

arkivanov:
Could you run this task locally? `./gradlew :sample-ios-app:build --stacktrace --info` may shed some light.

AlexeyKorshun:
it is success for me.

arkivanov:
I will check this tomorrow. Btw maybe we can just remove the `private var kittenView: KittenView? = nil` line at all? So just don't hold the KittenView, it can be a local variable in `viewDidLoad`.

arkivanov:
Can you try the proposed solution above? I thing this would be a better way to fix the retain loop. And maybe it will fix the build. 😀

AlexeyKorshun:
I will try it tomorrow :D

AlexeyKorshun:
@arkivanov 
> I will check this tomorrow. Btw maybe we can just remove the private var kittenView: KittenView? = nil line at all? So just don't hold the KittenView, it can be a local variable in viewDidLoad.
we can't do it, GC will remove Kitten view instance, because count of reference will be 0.

arkivanov:
> @arkivanov
> 
> > I will check this tomorrow. Btw maybe we can just remove the private var kittenView: KittenView? = nil line at all? So just don't hold the KittenView, it can be a local variable in viewDidLoad.
> 
> we can't do it, GC will remove Kitten view instance, because count of reference will be 0.

The KittenView reference is passed to the controller and is managed there.

AlexeyKorshun:
but if we remove it ?

arkivanov:
> but if we remove it ?

Sorry for confusion. I mean the KittenView reference is held by ViewController for no reason. It is not used by the ViewController itself. It is passed to KittenBinder via `kittenBinder.onViewCreated(view: kittenView!)` and is used and managed there. Am I missing something?

AlexeyKorshun:
right now I can't see code, will check all points tomorrow.

sample-ios-app/sample-ios-app/ViewController.swift
```diff
@@ -31,7 +31,7 @@ class ViewController: UIViewController {
     }
     
     private class KittenView: AbstractKittenView {
-        private let viewController: ViewController
+        private weak var viewController: ViewController?
```
arkivanov, line 34: So the `KittenBinder` accepts the `KittenView` and holds hard reference to it unless `onViewDestroyed` is called.

First thing we could fix is to not hold the `KittenView` in the `ViewController`:
`kittenBinder.onViewCreated(view: KittenView(viewController: self))`

After that there still will be a retain loop: `ViewController` -> `KittenBinder` -> `KittenView` -> `ViewController`. We can fixed this by destroying the `KittenBinder` not in `deinit` but in `viewWillUnload()`. The `KittenBinder` will clear the reference to the `KittenView`.

arkivanov (changes requested):

arkivanov:
The build should fixed now, please merge master to the branch 😄

sample-ios-app/sample-ios-app/ViewController.swift
AlexeyKorshun: viewWillUnload mark as [deprecated](https://developer.apple.com/documentation/uikit/uiviewcontroller/1621427-viewwillunload?language=objc)

AlexeyKorshun: I told with ios dev, and he said that we should use weak for reference to VC.

arkivanov: In our case the hard reference to view is cleared when the view is destroyed. What do you think about the following approach?
```swift
    private let kittenBinder = KittenBinder(storeBuilder: KittenStoreBuilderImpl())
    
    override func viewDidAppear(_ animated: Bool) {
        super.viewDidAppear(animated)
        kittenBinder.onViewCreated(view: KittenView(viewController: self))
        kittenBinder.onStart()
    }
        
    override func viewDidDisappear(_ animated: Bool) {
        kittenBinder.onStop()
        kittenBinder.onViewDestroyed()
        super.viewDidDisappear(animated)
    }
    
    deinit {
        kittenBinder.onDestroy()
    }
```

AlexeyKorshun: This solution is working, but should we recreate KittenView so often?

arkivanov: Is there any noticeable performance or memory impact? The KittenViee class itself is very lightweight, it is just a wrapper around the ViewConttoller.

AlexeyKorshun: I don't think that we have a problem with memory, but what is the reason why we need to recreate it? I can push changes with your approach today a nit later.

arkivanov: Well. Because we don't need to draw the view when it is disappeared. So we can actually release some memory when the ViewConttoller is invisible. And it also avoids memory leaks without `weak` and question marks.

arkivanov: Like in Android fragments, onDestroyView is called when the fragment is pushed to the back stack

AlexeyKorshun: yes but viewDidDisappear is the same like onPause or onStrop not onDestroyView and I think it is not the same :)

arkivanov: From Apple docs:

> Notifies the view controller that its view was removed from a view hierarchy.

So the view is removed from the hierarchy, is like pushing the fragment to the back stack.

Anyway, the KittenView is just a wrapper. The actual view is in the ViewController and is not recreated.

arkivanov (approved):

PR #561, opened 2020-11-13
Run linuxArm32Hfp tests on CI

### LINKED ISSUE #504: Run tests for linuxArm32Hfp target
The `linuxArm32Hfp` target can be built on any host, however running tests requires an ARM host. This is the reason why tests for this target are not running on CI currently. Looks like ARM is supported by GitHub Actions.

We need to investigate possible solutions and try to enable `linuxArm32Hfp` tests on CI.

Keep in mind, currently there is no corresponding Gradle task like `linuxArm32HfpTest`.

arkivanov:
Used the hint provided by @CherryPerry about `uraimo/run-on-arch-action` action.
Looks like this emulator is quite slow, so `AllBackgroundSchedulersTest` was hitting timeout. Updated the test. Other tests look fine.

Closes #504

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,6 +27,16 @@ jobs:
             ${{ matrix.os }}-gradle-
       - name: Validate Gradle Wrapper
         uses: gradle/wrapper-validation-action@v1
+      - name: Link linuxArm32Hfp
+        if: matrix.os == 'ubuntu-latest'
+        run: ./gradlew linkDebugTestLinuxArm32Hfp
+      - name: Test linuxArm32Hfp
+        if: matrix.os == 'ubuntu-latest'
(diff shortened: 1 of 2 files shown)


CherryPerry (approved):

CherryPerry:
@arkivanov 
I missed something. Check the logs:
> [==========] Running 6 tests from 3 test cases.
We run only 6 tests on ARM?

arkivanov:
This is very strange. There were lots of tests when I was testing. I will check.

CherryPerry:
> find . -name "test.kexe" -exec {}

This command finds all `kexe` files from all modules, right?

arkivanov:
Supposed to be. Looks like it runs only `reaktive-testing` module `kexe`. I will check why.

PR #562, opened 2020-11-13
Run linuxArm32Hfp tests per file

arkivanov:
Follow-up #561

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,7 +36,10 @@ jobs:
         with:
           arch: armv7
           distro: ubuntu20.04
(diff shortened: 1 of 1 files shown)


arkivanov:
I'm not sure what was the issue. When I was testing I used Ubuntu 18.04 as Doker distro, but then switched to Ubuntu 20.04. Maybe that "find" command behaves differently. Now switched to manual execution, file by file. I think this is even better. If output file names will change, the test will fail.

arkivanov:
Now its [fine](https://github.com/badoo/Reaktive/runs/1397961344?check_suite_focus=true).

CherryPerry (approved):

PR #577, opened 2021-01-15
Add middleware API

### LINKED ISSUE #574: Introduce a way to intercept operator assembly calls
Debugging errors with reaktive is a bit difficult right now if streams involve thread hopping because error stacktraces don't include traces from all threads on the JVM (and maybe native too?). Offering a way to intercept/hook into operator assembly calls may enable creation of something like [Traceur](https://github.com/T-Spoon/Traceur) for better stacktraces during development.

arkivanov:
Add middlewares for Reaktive sources wrapping.

Closes #574


arkivanov:
@CherryPerry @mk-jones @saket Here is a draft of middleware API. Could you please review? When this will be finalised for `Observable` I will expand it for other source types. Thanks!

arkivanov:
Sorry, will think a bit more about API.

saket:
I was surprised how small this PR was. What went wrong?

PR #578, opened 2021-01-15
Add plugins API

### LINKED ISSUE #574: Introduce a way to intercept operator assembly calls

arkivanov:
Add plugins API for Reaktive sources wrapping.

Closes #574


arkivanov:
@CherryPerry @mk-jones @saket Here is a draft of plugins API. Could you please review? When this will be finalised for Observable I will expand it for other source types. Also added usage example to the MainActivity. Thanks!

CherryPerry:
Sorry, I do not have any experience with RxJava plugins, so I can't review it from API client perspective.

arkivanov:
@CherryPerry  You can just provide your general thoughts about the API. The idea is to provide an ability to wrap all created sources with custom middlewares.

saket:
LGTM. Wanna add some documentation to `ObservablePlugins.onAssemble`?

reaktive/src/jvmJsCommonMain/kotlin/com/badoo/reaktive/observable/plugins/ObservablePlugins.kt
```diff
@@ -0,0 +1,11 @@
+package com.badoo.reaktive.observable.plugins
+
+import com.badoo.reaktive.annotations.ExperimentalReaktiveApi
+import kotlin.jvm.JvmField
+
+@ExperimentalReaktiveApi
+actual object ObservablePlugins {
+
+    @JvmField
```
CherryPerry, line 9: Make it also JvmStatic to remove singleton access?

reaktive/src/jvmJsCommonMain/kotlin/com/badoo/reaktive/observable/plugins/ObservablePlugins.kt
```diff
@@ -0,0 +1,11 @@
+package com.badoo.reaktive.observable.plugins
+
+import com.badoo.reaktive.annotations.ExperimentalReaktiveApi
+import kotlin.jvm.JvmField
+
+@ExperimentalReaktiveApi
+actual object ObservablePlugins {
```
CherryPerry, line 7: Implement also Maybe/Completable/Single?

arkivanov: Sure, right after we agree on the API

reaktive/src/jvmJsCommonMain/kotlin/com/badoo/reaktive/observable/plugins/ObservablePlugins.kt
arkivanov: Good catch!

sample-android-app/src/main/java/com/badoo/reaktive/sample/android/MainActivity.kt
```diff
@@ -12,12 +21,41 @@ class MainActivity : AppCompatActivity() {
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
 
+        ObservablePlugins.onAssemble =
```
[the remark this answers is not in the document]
arkivanov: My idea was to keep the API as simple as possible. With new "single" approach one still can create custom list of plugins as follows:
```
object MySinglePlugins : OnAssembleSingle {
    val onAssemble = ArrayList<OnAssembleSingle>()

    override fun <T> wrap(single: Single<T>): Single<T> =
        onAssemble.fold(single) { src, plugin -> plugin.wrap(src) }
}

fun onCreate() {
    SinglePlugins.onAssemble = MySinglePlugins
}
```

RxJava also provides similar API. But there is a drawback: it is not possible to have independent plugin libraries, so each adds its own plugins to the global list. Like with `Traceur`, it provides `Traceur.enableLogging()` method, which registers its internals via e.g. `RxJavaPlugins.setOnObservableAssembly`. Looks like you can use another similar "plugin" library, they will overwrite each other.

What do you think?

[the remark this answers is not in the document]
saket: Can plugins decorate/compose each other?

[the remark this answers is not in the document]
arkivanov: Yeah, looks like they can! I will think about an extension for this. Thanks!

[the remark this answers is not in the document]
arkivanov: @mk-jones 
1. Yes, something like this. Except ideally we should allow third-parties to add their plugins, instead of aggregating them on client side. To prevent cases like with Traceur, it just replaces any currently assigned plugin.
2. This was done intentionally, as I wanted to separate concerns. But I will think more about it. Thanks!

reaktive/src/jvmJsCommonMain/kotlin/com/badoo/reaktive/observable/plugins/ObservablePlugins.kt
arkivanov: So @JvmStatic can not be used together with @JvmField. And actually @JvmField in `object` makes it `static field`. So there is no singleton access on read/write.

sample-android-app/src/main/java/com/badoo/reaktive/sample/android/MainActivity.kt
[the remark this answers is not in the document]
arkivanov: @CherryPerry @mk-jones @saket Addressed all the comments. Could you please check again?

saket (dismissed):

mk-jones (dismissed):

reaktive/src/commonMain/kotlin/com/badoo/reaktive/plugin/ReaktivePluginsExt.kt
```diff
@@ -0,0 +1,26 @@
...
+        list = ArrayList()
+        plugins = list
+    }
+
+    list.add(plugin)
+}
+
+@ExperimentalReaktiveApi
+fun ReaktivePlugins.unregister(plugin: ReaktivePlugin) {
+    plugins?.remove(plugin)
```
CherryPerry, line 19: Check plugins size and null-out reference?

reaktive/src/commonMain/kotlin/com/badoo/reaktive/plugin/SimpleReaktivePlugin.kt
```diff
@@ -0,0 +1,10 @@
+package com.badoo.reaktive.plugin
+
+import com.badoo.reaktive.annotations.ExperimentalReaktiveApi
+import com.badoo.reaktive.observable.Observable
+
+@ExperimentalReaktiveApi
+open class SimpleReaktivePlugin : ReaktivePlugin {
```
CherryPerry, line 7: What is the point of this class?
Maybe better to provide default implementation of each method in `ReaktivePlugin`?

reaktive/src/commonMain/kotlin/com/badoo/reaktive/plugin/ReaktivePluginsExt.kt
arkivanov: Will do

reaktive/src/commonMain/kotlin/com/badoo/reaktive/plugin/SimpleReaktivePlugin.kt
arkivanov: For some reason I don't like methods with implementations in interfaces. They may cause confusion/conflicts/bugs with multiple/nested inheritance. The purpose of this class is you can extend it and implement required methods. Do you think in this particular case it would be good to add default implementations in interface?

CherryPerry: You can keep it as is if you want.

CherryPerry (dismissed):

reaktive/src/commonMain/kotlin/com/badoo/reaktive/plugin/ReaktivePluginsExt.kt
CherryPerry: Implement it? We can avoid `fold` invocation on every call after unregistering all plugins.

CherryPerry (changes requested):

reaktive/src/commonMain/kotlin/com/badoo/reaktive/plugin/ReaktivePluginsExt.kt
arkivanov: Oops, forgot about it! Thanks I will fix!

arkivanov:
Native tests are crashing, I will fix

arkivanov:
@CherryPerry I added two commits:
1. Converted plugins API to "top level style", removed `object ReaktivePlugins` and replaced with top level properties and functions. This is to comply with the existing API, like `reaktiveUncaughtErrorHandler`, `resetReaktiveUncaughtErrorHandler()`, `overrideSchedulers(...)`, etc.
2. Fixed the native SIGSEGV crash. This was due to the "undefined" property initialisation order in Kotlin, when in properties are in different files. Looks like (not only?) `SharedImmutable` properties are initialised at startup, and our various top-level `val xxxOfEmpty` properties were calling `onAssembleXxx`, which in turn were calling the unitialised (yet) top-level property `plugins`.

CherryPerry (approved):

PR #597, opened 2021-04-09
Add SerialDisposable

nikolaykasyanov:
Hey, first of all, thanks for this awesome library.

I thought it'd be nice to have this type of disposable inspired by [RxSwift][rxswift] mostly, however I'm open to API and implementation suggestions/guidance.

[rxswift]: https://github.com/ReactiveX/RxSwift/blob/b4e4ea9b238a868b40b4951434f38a25c6c1b8ab/RxSwift/Disposables/SerialDisposable.swift

### DIFF
--- a/reaktive/src/commonMain/kotlin/com/badoo/reaktive/disposable/SerialDisposable.kt
+++ b/reaktive/src/commonMain/kotlin/com/badoo/reaktive/disposable/SerialDisposable.kt
@@ -1,0 +1,45 @@
+package com.badoo.reaktive.disposable
+
+import com.badoo.reaktive.utils.atomic.AtomicReference
+import com.badoo.reaktive.utils.atomic.update
+
+/**
+ * Disposable whose underlying disposable can be replaced, causing this underlying resource's disposal.
+ *
+ * When [SerialDisposable] is disposed, it's underlying disposable is disposed and all disposables assigned
+ * after that are disposed immediately.
+ */
+class SerialDisposable : Disposable {
+    private val current = AtomicReference<Disposable?>(Disposable())
+
+    /**
+     * Contains the current disposable. If none has been set or [isDisposed], returns an empty disposable.
+     * Disposes the current disposable upon setting.
+     *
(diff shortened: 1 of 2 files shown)


reaktive/src/commonMain/kotlin/com/badoo/reaktive/disposable/SerialDisposable.kt
```diff
@@ -0,0 +1,45 @@
...
+import com.badoo.reaktive.utils.atomic.update
+
+/**
+ * Disposable whose underlying disposable can be replaced, causing this underlying resource's disposal.
+ *
+ * When [SerialDisposable] is disposed, it's underlying disposable is disposed and all disposables assigned
+ * after that are disposed immediately.
+ */
+class SerialDisposable : Disposable {
+    private val current = AtomicReference<Disposable?>(Disposable())
```
nikolaykasyanov, line 13: I noticed that `CompositeDisposable` is implemented with `synchronized` on JVM, unlike on native, please let me know if I should do the same here.

arkivanov:
Hi, thanks for the PR! Isn't it the same as the [DisposableWrapper](https://github.com/badoo/Reaktive/blob/master/reaktive/src/commonMain/kotlin/com/badoo/reaktive/disposable/DisposableWrapper.kt)?

reaktive/src/commonMain/kotlin/com/badoo/reaktive/disposable/SerialDisposable.kt
arkivanov: Native `CompositeDisposable` is synchronized using atomics, same as `DisposableWrapper`.

nikolaykasyanov:
Yeah looks like it, could find it due to unusual naming. Gonna close this one then, thanks!

arkivanov:
> Yeah looks like it, could find it due to unusual naming. Gonna close this one then, thanks!

Thanks for the feedback, I will consider renaming.

arkivanov:
Filed #598

PR #607, opened 2021-05-07
Introduce SerialDisposable, deprecate DisposableWrapper

### LINKED ISSUE #598: Consider renaming DisposableWrapper to SerialDisposable
To follow the naming from all other Rx libraries.

arkivanov:
Closes #598


reaktive/src/commonMain/kotlin/com/badoo/reaktive/disposable/DisposableWrapper.kt
```diff
@@ -3,6 +3,7 @@ package com.badoo.reaktive.disposable
 /**
  * Thread-safe container of one [Disposable]
  */
+@Deprecated("Please use SerialDisposable", ReplaceWith("SerialDisposable"))
 @Suppress("EmptyDefaultConstructor")
 expect open class DisposableWrapper() : Disposable {
```
CherryPerry, line 8: `@Deprecated open class DisposableWrapper(): SerialDisposable()`

So you do not need to copy-paste implementation and still have backward compatibility.

CherryPerry (changes requested):

reaktive/src/commonMain/kotlin/com/badoo/reaktive/disposable/DisposableWrapper.kt
arkivanov: Yeah, I did it first, but then decided it will be a breaking change. Now I see it won't: https://wiki.eclipse.org/Evolving_Java-based_APIs_2#Evolving_API_Classes
Am I right it will not break binary compatibility?

CherryPerry: It won't. The signature of methods is the same. Also the class still implements `Disposable` interface. So should be fine.

arkivanov: Fixed, thanks!

CherryPerry (approved):

Issue #609, opened 2021-05-12
Update Kotlin to 1.5.0


arkivanov:
What's new: https://kotlinlang.org/docs/whatsnew15.html

PR #610, opened 2021-05-14
Update kotlin to 1.5.0

arkivanov:
Closes https://github.com/badoo/Reaktive/issues/609


utils/src/jsMain/kotlin/com/badoo/reaktive/utils/ThrowableExt.kt
```diff
@@ -1,5 +1,5 @@
 package com.badoo.reaktive.utils
 
 actual fun Throwable.printStack() {
-    console.log(this)
+    console.log(toString())
```
arkivanov, line 4: The `log` function now throws an exception if a non-string object is passed.

benchmarks/jmh/src/jmh/kotlin/com/badoo/reaktive/benchmarks/jmh/scrabble/ScrabbleBase.kt
```diff
@@ -50,7 +50,7 @@ abstract class ScrabbleBase {
 
     private fun readResource(path: String) =
         BufferedReader(InputStreamReader(GZIPInputStream(this.javaClass.classLoader.getResourceAsStream(path)))).lines()
-            .map { it.toLowerCase() }.collect(Collectors.toSet())
+            .map { it.lowercase() }.collect(Collectors.toSet())
```
arkivanov, line 53: `toLowerCase`/`toUpperCase` are now deprecated. Could not update the similar code in the `DarwinPlugin`, it does not see the new function and the compilation fails.

includedBuild/gradleConfiguration/src/main/kotlin/com/badoo/reaktive/configuration/MppConfigurationPlugin.kt
```diff
@@ -39,6 +40,21 @@ class MppConfigurationPlugin : Plugin<Project> {
...
+                project.tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile>().configureEach {
+                    kotlinOptions {
+                        /*
+                         * Starting from Kotlin 1.5.0 the default jvmTarget is 1.8, and the 1.6 target is deprecated.
+                         * For some reason some JVM tests are failing with
+                         * java.lang.NoSuchMethodError: 'int java.lang.Integer.compareTo(int)'.
+                         * The Integer.compareTo method was added in Java 1.8.
+                         * We should find a fix, or avoid using the Integer.compareTo method.
+                         */
+                        jvmTarget = "1.6"
```
CherryPerry, line 53: I did not get this part.
They are failing because of Java 1.8?

benchmarks/jmh/src/jmh/kotlin/com/badoo/reaktive/benchmarks/jmh/scrabble/ScrabbleBase.kt
CherryPerry: Because for Gradle plugins I use built-in Kotlin version instead of the latest one. For `gradle-6.8.3` it is 1.4.

includedBuild/gradleConfiguration/src/main/kotlin/com/badoo/reaktive/configuration/MppConfigurationPlugin.kt
arkivanov: I'm not sure why does it fail. I suspect that tests are somehow compiled with 1.6 and so the method is not found. Also not sure why it does not fail in master. Could not google anything. Will appreciate any help here.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.5-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.8.3-all.zip
```
CherryPerry, line 3: Please update Gradle wrapper properly via `gradlew wrapper --gradle-version=6.8.3 --distribution-type=all`.

includedBuild/gradleConfiguration/src/main/kotlin/com/badoo/reaktive/configuration/MppConfigurationPlugin.kt
CherryPerry:
```kotlin
private val comparator = Comparator<Int> { a, b ->  a.compareTo(b) }
```

It fixes the issue. I do not know why method reference does not work, might be a new IR backend bug related to method referencing.

arkivanov: Interesting, I did not even try since the error message is pretty much descriptive. Will apply, thanks!

arkivanov: Filed [KT-46699](https://youtrack.jetbrains.com/issue/KT-46699)

CherryPerry (approved):

PR #611, opened 2021-05-14
binary-compatibility-validator 0.5.0 fork

CherryPerry:
Fork binary-compatibility-validator 0.5.0 and enable dump only for jvm target.

Most of the changes are just `git clone`. Did not want to fork it as a separate repo on GitHub.


tools/binary-compatibility-validator/src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt
```diff
@@ -0,0 +1,166 @@
...
+            }.all {
+                project.configureKotlinCompilation(it, extension, useOutput = true)
+            }
+        }
+
+        project.pluginManager.withPlugin("kotlin-multiplatform") {
+            if (project.name in extension.ignoredProjects) return@withPlugin
+            val kotlin = project.extensions.getByName("kotlin") as KotlinMultiplatformExtension
+            kotlin.targets.matching {
+                it.platformType == KotlinPlatformType.jvm
```
CherryPerry, line 62: Removed `KotlinPlatformType.jvm` here.

arkivanov:
Seems like we should comply with the Apache license's points `4.a` and `4.b`: https://github.com/Kotlin/binary-compatibility-validator/blob/master/LICENSE.TXT
Perhaps copy LICENSE.txt to the folded and also add some comments to the modified files.

arkivanov (changes requested):
See the comment above

CherryPerry:
Added notice e2d8047, LICENSE file is already provided.

arkivanov (approved):

PR #614, opened 2021-05-21
Added maxConcurrency argument to flatMap

### LINKED ISSUE #612: Recommended way to specify maximum concurrency?
Is there a way to specify maximum threads concurrency when using Reaktive?

For Rx I would use flatmap and use the maxConcurrency parameter (see https://github.com/politrons/reactive/blob/master/src/test/java/rx/observables/transforming/ObservableFlatMap.java)

arkivanov:
Closes #612 
Related docs: http://reactivex.io/RxJava/javadoc/io/reactivex/Observable.html#flatMap-io.reactivex.functions.Function-boolean-int-

### DIFF
--- a/reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/FlatMap.kt
+++ b/reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/FlatMap.kt
@@ -19,7 +19,7 @@ fun <T, R> Observable<T>.flatMap(mapper: (T) -> Observable<R>): Observable<R> =
     flatMap(maxConcurrency = Int.MAX_VALUE, mapper = mapper)
 
 fun <T, R> Observable<T>.flatMap(maxConcurrency: Int, mapper: (T) -> Observable<R>): Observable<R> {
-    require(maxConcurrency > 0) { "masConcurrency value must be positive" }
+    require(maxConcurrency > 0) { "maxConcurrency value must be positive" }
 
     return observable { emitter ->
         val upstreamObserver = FlatMapObserver(emitter.serialize(), maxConcurrency, mapper)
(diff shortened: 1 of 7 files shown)


reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/FlatMap.kt
```diff
@@ -53,12 +91,56 @@ private class FlatMapObserver<in T, in R>(
...
+    private val refCounter = RefCounter(lock::destroy)
+    private val count = AtomicInt(limit)
+    private val queue = SharedQueue<T>()
+
+    private val _isDisposed = AtomicBoolean(false)
+    override val isDisposed: Boolean get() = _isDisposed.value
+
+    override fun dispose() {
+        if (_isDisposed.compareAndSet(expectedValue = false, newValue = true)) {
+            sync(queue::clear)
```
mk-jones, line 115: What would happen in the case that FlatMapQueue .offer(...) just after sync(queue::clear)? It looks like it would invoke the callback with the new value despite being disposed. 
However, I don't think is a problem as the subscribe call will check the disposed state.

arkivanov: The callback might be called under race condition, which will call `subscribeInner`, which subscribe the inner `Observable`. The latter will emit `Disposable`, which will be immediately disposed. Kinda same behaviour exist even without `maxConcurrency`, when upstream emits an item while the downstream is being disposed.

I updated the disposing logic inside `FlatMapQueue`, so now all resources are cleared inside the `RefCounter` block, not only the `Lock`. This won't prevent the concern, but it will be ensure there are no items left in the `Queue` at the very end.

mk-jones (approved):

PR #619, opened 2021-06-11
Added watchosX64 target

### LINKED ISSUE #617: Support watchosX64 target
Also looks like we have duplicated `watchosArm32Abc` source sets.

cc: @CherryPerry

arkivanov:
Closes #617

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,5 @@
 buildscript {
-    ext.reaktive_version = '1.1.22'
+    ext.reaktive_version = '1.1.23-t1'
     ext.reaktive_group_id = 'com.badoo.reaktive'
 }
 
--- a/build.gradle
+++ b/build.gradle
@@ -1,5 +1,5 @@
 buildscript {
-    ext.reaktive_version = '1.1.23-t1'
+    ext.reaktive_version = '1.1.23'
     ext.reaktive_group_id = 'com.badoo.reaktive'
 }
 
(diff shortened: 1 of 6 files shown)


build.gradle
```diff
@@ -1,5 +1,5 @@
 buildscript {
-    ext.reaktive_version = '1.1.22'
+    ext.reaktive_version = '1.1.23-t1'
```
CherryPerry, line 2: Why `-t1`?

includedBuild/dependencies/src/main/kotlin/com/badoo/reaktive/dependencies/Deps.kt
```diff
@@ -3,7 +3,7 @@ package com.badoo.reaktive.dependencies
 object Deps {
 
     private const val kotlinVersion = "1.5.0"
```
CherryPerry, line 5: Also we can upgrade to 1.5.10.

build.gradle
arkivanov: Accidentally, was checking the publication, thanks for spotting!

includedBuild/dependencies/src/main/kotlin/com/badoo/reaktive/dependencies/Deps.kt
arkivanov: Done

CherryPerry (approved):

Issue #626, opened 2021-06-29
Can reaktive catch errors thrown by nested coroutines?


saket:
I’m debugging a case where an exception thrown inside a nested coroutine isn’t bubbling up when singleFromCoroutine is used. I think I can try catch all my nested coroutines to solve this, but can reaktive’s coroutines interop help here by using a `CoroutineExceptionHandler` for launching coroutines?

My understanding of error handling in coroutines comes from this article: https://www.lukaslechner.com/why-exception-handling-with-kotlin-coroutines-is-so-hard-and-how-to-successfully-master-it/

arkivanov:
Thanks! I think the error handling can be improved. I will check.

arkivanov:
Closed via #628 and #630

PR #630, opened 2021-07-09
Use CoroutineExceptionHandler in coroutines-interop nmtc

arkivanov:
Same as #628 but for `mt-coroutines` branch.
The last part for #626.


arkivanov:
Since the changes are mostly copied from #628, no need to review.

Issue #398, opened 2020-02-02
Document the code with KDoc


arkivanov:

arkivanov:
- [ ] Observable
- [ ] Maybe
- [X] Single
- [ ] Completable
- [ ] Subjects

PR #632, opened 2021-07-30
Added KDocs for Maybe, part 1

arkivanov:
As part of #398


arkivanov:
Alright, since it is mostly copy-pasted KDocs from `Single` and `Completable`, I think we can proceed without review.

Issue #488, opened 2020-06-23
Publish Reaktive node.js target to npm


minaEweida:
Is there any plans to publish Reaktive to npm? 
Recently we used reaktive in our lib to support node.js apps so we are publishing Reaktive to our internal artifactory since Reaktive is not published

arkivanov:
Thanks @minaEweida we'll check.

minaEweida:
Hi @arkivanov, if you are planning to do this, please make sure to mention that supported node.js version should be 12+ since `MainScheduler` is using `globalThis` which is supported for Node.js version 12.

<img width="1201" alt="Screen Shot 2020-07-21 at 3 00 12 PM" src="https://user-images.githubusercontent.com/20396931/88047596-e9533780-cb62-11ea-8ab4-3ae7d9d65914.png">

arkivanov:
A related ticket: [KT-41566](https://youtrack.jetbrains.com/issue/KT-41566)

arkivanov:
NPM publishing plugin is released: https://gitlab.com/lt.petuska/npm-publish/-/releases/v0.1.1

arkivanov:
Looks like a good article about publishing to npm: https://zellwk.com/blog/publish-to-npm/

arkivanov:
Maybe anybody want to contribute? We could not find anyone with experience in both Kotlin Multiplatform and NPM at the same time. From our side we could setup everything needed (like registering with the NPM repository, etc.)

minaEweida:
We managed to publish our lib on our internal NPM and as a workaround we published reaktive artifacts as well. I will try to do the same here.

gciatto-unibo:
NPM publishing would be very nice for my as it would enable its exploitation in my Kotlin [MP project](https://github.com/tuProlog/2p-kt).

BTW I'm well aware of the issues related to publishing Kotlin MPP to NPM.
I've implemented a simple plugin to simplify the process: https://github.com/gciatto/kt-npm-publish

It is not clear to me what's the limiting factor in this case, but I hope my plugin may help.

arkivanov:
@gciatto-unibo The limiting factor here is the lack of expertise in this area. We will need to learn all the things first, hence the delay. As mentioned above we are happy to accept contributions for this task. And we will assist as much as we can.

gciatto-unibo:
I have some degree of expertise concerning how to publish Kotlin-JS stuff on NPM.
Let me know if and how I can help.

[...]

PR #635, opened 2021-08-25
npm-publish example

mpetuska:
A POC setup of npm-publish with a local consumer.

# Testing
0. Install yarn
1. `./gradlew :reaktive:pack` - build local npm distribution
2. `cd reaktive/ts-consumer` - get to npm module
3. `yarn install` - install npm dependencies
4. `yarn start` - run the sample script that uses reaktive
5. `yarn build` - build js output from ts
6. `yarn reinstall` - reinstall local `reaktive` library (if you rebuild it with changes on gradle)
7. `yarn rebuild` - rebuild reaktive gradle module and reinstall it to this consumer (recommended to run after any kt changes)

# Notes
Most of the things work out of the box, however there are few caveats
* Turns out kotlin stdlib is not bundled by IR (good!), so I had to add that to package.json explicitly (next version of npm-publish plugin will fix that). EDIT: This is incorrect, stdlib IS bundled afterall, tsc just complains because it has no types. Resulting ts-consumer js output is valid and runs fine.
* Anything that you want visible for js consumer MUST be annotated with `@kotlin.js.JsExport` annotation, which imposes its own [restrictions](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.js/-js-export/). For that I've created a new sandbox file to showcase usage (reaktive/src/jsMain/kotlin/com/badoo/reaktive/npmPublishTestSandbox.kt)
* Packed distribution lives in `reaktive/build/publications/npm` if you want to inspect assembled output

Relates to #488


arkivanov:
@mpetuska Could you explain in more detail why do we need the local consumer?

reaktive/src/jsMain/kotlin/com/badoo/reaktive/npmPublishTestSandbox.kt
```diff
@@ -0,0 +1,18 @@
+package com.badoo.reaktive
+
+import kotlin.js.Promise
+
+@JsExport
```
arkivanov, line 5: Can we instead of this sample code being put in production folder, just js-export some real Reaktive things? E.g. the `Observable` interface and all its super types. To check the whole use case, we can create the following class in `sample-mpp-module` module:

```kotlin
@JsExport
class ExportedThing {
    val data: Observable<String> = observableOf("one", "two", "three")
}
```

Then we could also npm-publish the sample module and check how it can be used in a JS project. Would this work/help?

mpetuska:
Oh this is not to be merged. It's just to showcase you guys how the setup might look like. Ts-consumer is entirely optional, but i find it useful to test published version of npm module before you publish

reaktive/build.gradle.kts
```diff
@@ -0,0 +1,54 @@
+plugins {
```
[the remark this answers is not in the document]
mpetuska: Nope, I'm just more comfortable with it.

reaktive/src/jsMain/kotlin/com/badoo/reaktive/npmPublishTestSandbox.kt
mpetuska: I tried, but it's quite a rabbit hole once you start annotating the trail, so I decided to leave that to you guys, since you know your domain best.

arkivanov:
Thank you @mpetuska for the showcasing PR! It will be very useful for us. Since it is not intended to be merged, I'm closing it now. I will create a separate PR based on this one.

Issue #523, opened 2020-08-18
Enable HMPP


arkivanov:
Would be nice to enable, but currently blocked by [KT-41160](https://youtrack.jetbrains.com/issue/KT-41160).

omainegra:
Could it be possible to use the workaround suggested in the ticket in the meantime?

arkivanov:
I will try. But I remember there were more compilation errors. Also according to multiple reports this feature is really buggy and looks unfinished. Personally I could not enable it in any of my projects.

arkivanov:
Just in case, sharing code between targets is possible for a long time already. There is just no IDE support for this, but there are workarounds though.

omainegra:
It's bad to hear that, I'm migrating one library project and it's actually working, and the build script is way simpler now, just for publishing I'm disabling it

arkivanov:
Yeah, I understand. I will try one more time, but most likely it won't work at the moment.

arkivanov:
Fol a library development it should not be a blocker, since code sharing is possible in any intermediate source set. E.g. jvmCommonMain can even depend on both jvmJsCommonMain and jvmNativeCommonMain at the same time.

arkivanov:
IDE can be temporary made happy with a workaround

arkivanov:
Just tried one more time. Another blocker is that with HMPP enabled we can't access `platform.*` in the `nativeCommonMain` source sets (in our case it is shared between all Linux and Darwin targets). E.g. `platform.posix.fprintf` does not compile. We have plenty of code there. In order to compile this with HMPP we need to duplicate this for Linux and Darwin. And perhaps Darwin should be also split (not sure yet). This significantly increases amount of code (duplicated one) and makes it harder to maintain. Failed the issue: [KT-42574](https://youtrack.jetbrains.com/issue/KT-42574).

At the moment it is better to wait for improvements in HMPP.

arkivanov:
But let's collect use cases! It should be possible to use a library published without HMPP in a project with HMPP enabled, and access its common code from any HMPP shared source set. Only library code from a shared source set might not be accessible.

omainegra:
Thanks for the efforts, in my case it's completely fine to use HMPP and still use Reaktive, the only thing I have to do since my project is also a library is disabled it during publish (`./gradlew -Pkotlin.mpp.enableGranularSourceSetsMetadata=false publish`)

arkivanov:
The following issues are still relevant:
[...]

PR #638, opened 2021-09-17
Prepare for HMPP

arkivanov:
This PR prepares the project for [HMPP](https://kotlinlang.org/docs/mpp-share-on-platforms.html).

Currently our CI uses two hosts: macOS and Ubuntu. We disable Darwin targets when building on Ubuntu, and disable non-Darwin targets when building on macOS. Disabling targets does not work well with HMPP, it is encouraged to always have all targets enabled. CI optimisations can be done by either disabling unnecessary Gradle tasks, or by calling only required tasks from CI. I chose the former option, as it seems that calling just `build` task from CI is less error-prone.

The following issues are still relevant, we should wait for fixes before we could enable HMPP:
- [KT-41509](https://youtrack.jetbrains.com/issue/KT-41509)
- [KT-48811](https://youtrack.jetbrains.com/issue/KT-48811)
- [KT-43309](https://youtrack.jetbrains.com/issue/KT-43309) (should be fixed by [KT-48288](https://youtrack.jetbrains.com/issue/KT-48288) in Kotlin 1.6)

As part of #523


arkivanov:
@sellmair It would be awesome if you could review this PR as well! Your help would be much appreciated.

sample-js-browser-app/build.gradle
```diff
@@ -1,9 +1,12 @@
+import com.badoo.reaktive.configuration.Target
 import com.badoo.reaktive.dependencies.Deps
 
-plugins {
-    id 'org.jetbrains.kotlin.js'
+if (!Target.shouldDefineTarget(project, Target.JS)) {
+    return
 }
 
+plugins.apply("org.jetbrains.kotlin.js")
```
CherryPerry, line 8: I think having a list of modules to compile on macOS is better than using `apply` instead of `plugins` as autocompletion might not work properly in the script.

CherryPerry:
In general I do not see any profit from this PR except compliance with new publication scheme. The amount of code increased due to all of this KT bug and I we still need to define a lot of dependencies for a lot of source sets.

arkivanov:
> In general I do not see any profit from this PR except compliance with new publication scheme. The amount of code increased due to all of this KT bug and I we still need to define a lot of dependencies for a lot of source sets.

The main reason for this PR is that HMPP will be the default mode in Kotlin 1.6 with an opt-out flag. The opt-out flag will be also removed at some point later.

arkivanov:
> In general I do not see any profit from this PR except compliance with new publication scheme. The amount of code increased due to all of this KT bug and I we still need to define a lot of dependencies for a lot of source sets.

Also the workarounds (the code duplication) should go away eventually. E.g. [KT-48288](https://youtrack.jetbrains.com/issue/KT-48288) should be fixed in Kotlin 1.6, which will remove ~50% of the workarounds.

sample-js-browser-app/build.gradle
arkivanov: My intention was to unify build/yaml scripts for consistency (e.g. we have similar logic in `sample-ios-app` and `sample-macos-app` modules), and also we might add Windows targets in future. But if you believe the current approach is better, I will revert the changes. Thanks!

CherryPerry: Revert please, we do not have too much modules to maintain and we are more explicit this way.

sellmair:
@arkivanov Is correct: We are planning to switch to HMPP by default soon. The opt-out flag will not be supported forever, but for "quite some time" ™ definitely. We (the multiplatform-core) team also offer our help here with any further issues. 

I checked out this branch from @arkivanov and tested it with features planned to ship in 1.6.20. In this current state it seems that most (if not all 🤞) workarounds for missing commonized functions can be removed. I will absolutely keep an eye on this project and will test further changes on this project.

arkivanov:
Thanks everyone for your inputs! I think we should merge the changes of how we disable targets, which I introduced in this PR - so all targets are defined at all times, and unnecessary tasks are switched off. But we can revert all the changes in the production code and disable HMPP for now. We can check things later again after Kotlin `1.6` and/or `1.6.20`. Once there are fewer (or no) workarounds, we could finally enable HMPP.

arkivanov:
@CherryPerry I reverted all the changes in production code and disabled HMPP for now. Only build configuration is changed.

CherryPerry (approved):

Issue #608, opened 2021-05-08
Update binary compatibility plugin


CherryPerry:
The latest version – 0.5.0.

CherryPerry:
Blocked by https://github.com/Kotlin/binary-compatibility-validator/issues/47

arkivanov:
Tried to update Kotlin version to `1.5.0` and the `binary-compatibility-validator` now dumps `internal` things. Could it be related?
See #610

CherryPerry:
I think it might be.
You can try follow instructions in my comment in the related issue (disable `jvm` or `android` target and upgrade Kotlin version) and see what will happen.

arkivanov:
If I understood correctly, I updated Kotlin to `1.5.0` and disabled `// setupAndroidTarget(project)` - internal things are still being dumped.

CherryPerry:
And upgraded version of compatibility plugin to 0.5.0?

arkivanov:
Looks like related to https://github.com/Kotlin/binary-compatibility-validator/issues/54 , we should to update try the new version of the plugin

arkivanov:
Nah, then we will hit this issue :-) Looks like the update to `1.5.0` is also _kinda_ blocked by this task.

CherryPerry:
I can make a patch to dump only `android` jar. Looks pretty simple [here](https://github.com/Kotlin/binary-compatibility-validator/blob/master/src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt#L62).

arkivanov:
I can try, but it might be dangerous, we can miss something JVM-specific. Maybe better to keep internal things and just check manually per case? Not sure

CherryPerry:
Actually in the current version of the plugin only `jvm` target is dumped [here](https://github.com/Kotlin/binary-compatibility-validator/blob/115f26ff22/src/main/kotlin/BinaryCompatibilityValidatorPlugin.kt#L62).

arkivanov:
Then perhaps we can disable `android`, not `jvm`?

CherryPerry:
Anyway, can you check if dump works fine with validator 0.5.0 + kotlin 1.5.0 + disabled `android` sourceset on your branch? Meanwhile I will prepare separate PR with validator.

arkivanov:
Kotlin `1.5.0` and plugin `0.5.0` work fine with the `android` target disabled.

CherryPerry:
[...]

PR #647, opened 2021-10-30
Switch to official compatibility validator plugin

emartynov:
## Task
Switch from the for to official JetBrain plugin - Fix for #608

## Changes
* Remove for of the plugin
* Update official plugin version

## Notes
* Not sure where is magic - I didn't find why removal of fork automatically switch to official plugin. Also according to [official documentation](https://github.com/Kotlin/binary-compatibility-validator) I have to change plugin id. However, this broke the gradle sync. So I kept plugin id as it is and command line just worked fine.


CherryPerry:
Can you remove existing `api` folders and create new dump using [apiDump](https://github.com/Kotlin/binary-compatibility-validator#tasks) please? Looks like new version has new `api` folder structure, so `apiCheck` on CI check will fail.

> Not sure where is magic

[Included build](https://docs.gradle.org/current/userguide/composite_builds.html) replaces remote artefacts when has the same publication id and version.

emartynov:
I hope, I did everything correct.

CherryPerry (approved):

emartynov:
Don't you people squash PRs?

CherryPerry:
By default – no, your commit history looks fine enough.

Issue #620, opened 2021-06-11
[2.0.0] API improvements and other fixes


arkivanov:
There are some APIs that can be improved/simplified. This issue is to collect all the cases. This is also a good place to raise and discuss any proposals or concerns.

- [x]  `flatMap` - add default value for the `maxConcurrency` argument, remove `flatMap` variants without the `maxConcurrency` argument
- [x]  `flatMap` with `resultSelector` - reorder arguments in the following way: `resultSelector`, `maxConcurrency`, `mapper`
- [x]  `observableInterval` - put `startDelayMillis` after `periodMillis` and add default value `startDelayMillis: Long = periodMillis`, remove another `observableInterval` variant
- [x]  `Observable.replay` - add default for `bufferSize` argument, remove the variant without arguments
- [x] `Completable.asMaybe` and `Completable.asObservable` - remove the generic parameter and make the return generic type `Nothing`
- [x] `doOnAfterSuccess` - rename `action` argument to `consumer`
- [x]  remove all deprecated code
- [x]  `repeat(count: Int = -1)` - the argument name should be `times`, for consistency with `retry` and RxJava. The type should be `Long`, and the default value should be `Long.MAX_VALUE`.
- [x]  `repeat` with value `0` should return empty `Observable`, `repeat` with value `1` should subscribe only once, and so on. For consistency with RxJava.
- [x]  `retry(times: Int)` - the argument type should be `Long`, and the default value should be `Long.MAX_VALUE`.
- [x]  `Observable.scan(getSeed, accumulate)` and `Observable.scan(accumulate)`, - the `accumulate` argument should be renamed to `accumulator` for consistency with RxJava.
- [x]  `Observable.throttle` - should accept a `Scheduler` for timings instead of using `Clock` implicitly, for better testing on the client side.
- [x]  Review all `ExperimentalReaktiveApi`, perhaps we can stabilise something.
- [x]  `collect` should have a mutable collection supplier argument, `accumulator` function should return `Unit`.
- [x] Convert all `millis` durations to the new `Duration` API, if it becomes stable
- [ ] ~~Wait for Multiple Receivers to become stable and extract all `subscribeXyz` extensions from `DisposableScope`, and also `T.scope(dispose: (T) -> Unit)`~~ (removed, context receivers will become stable no earlier than Kotlin 2.0)
- [x] Stabilize the Plugins API
- [x] Remove the old memory model support, optimize for the new memory model
[...]

PR #649, opened 2021-10-30
Remove deprecated API

emartynov:
## Task
Remove deprecated API part of #620


reaktive/src/commonTest/kotlin/com/badoo/reaktive/completable/CompletableToCompletableTests.kt
```diff
@@ -23,9 +23,6 @@ class CompletableToCompletableTestsImpl(
     transform: Completable.() -> Completable
 ) : CompletableToCompletableTests, SourceTests by SourceTestsImpl(TestCompletable(), { transform().test() }) {
 
-    @Deprecated("Just to fix complilation issues")
-    constructor() : this({ throw UnsupportedOperationException() })
```
arkivanov, line 27: Hmm, it's not part of the public API. But there were compilation errors a while ago. Does it compile across all targets?

arkivanov:
Also could you please run apiDump Gradle task and commit the updated api files? Otherwise build will fail.

arkivanov:
Also could you please run apiDump Gradle task and commit the updated api files? Otherwise build will fail.

reaktive/src/commonTest/kotlin/com/badoo/reaktive/completable/CompletableToCompletableTests.kt
emartynov: Not sure, let me check. The JVM was compiling fine.

emartynov:
@arkivanov dump is updated and CI looks green. Not sure what compilation error it was before.

arkivanov (approved):

arkivanov:
Thanks for you contribution!

PR #650, opened 2021-10-31
[2.0.0] Retry and repeat changes

emartynov:
## Task
Change type and rename parameters for Retry and Repeat #620.

## Changes
* Rename `count` to `times` for `Repeat` and make it Long. Also change default parameter value to `Long.MAX_VALUE`.
* Change type to long for `Retry` (see notes)
* Small changes around tests

## Notes
* Unfortunately if I put default value for retry we have clashing function calls. If my knowledge of Kotlin is correct, then I have to provide some parameter to help compiler to resolve type. This might be inconvenient for the user of the library. However, version with default selector is less performant. We could remove there default parameter value and use the simpler instead.

### DIFF
--- a/reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/Repeat.kt
+++ b/reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/Repeat.kt
@@ -17,7 +17,7 @@ fun <T> Observable<T>.repeat(times: Long = Long.MAX_VALUE): Observable<T> {
     return observable { emitter ->
         val observer =
             object : ObservableObserver<T>, ValueCallback<T> by emitter, ErrorCallback by emitter {
-                private val counter: AtomicLong? = if (times >= 0) AtomicLong(times) else null
+                private val counter: AtomicLong? = if (times < Long.MAX_VALUE) AtomicLong(times) else null
 
                 // Prevents recursive subscriptions
                 private val serializer =
--- a/reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/Repeat.kt
+++ b/reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/Repeat.kt
@@ -4,18 +4,19 @@ import com.badoo.reaktive.base.ErrorCallback
 import com.badoo.reaktive.base.ValueCallback
 import com.badoo.reaktive.disposable.Disposable
 import com.badoo.reaktive.utils.atomic.AtomicInt
+import com.badoo.reaktive.utils.atomic.AtomicLong
 import com.badoo.reaktive.utils.serializer.serializer
 
 /**
- * Returns an [Observable] that automatically resubscribes to this [Observable] at most [count] times.
+ * Returns an [Observable] that automatically resubscribes to this [Observable] at most [times] times.
  *
  * Please refer to the corresponding RxJava [document](http://reactivex.io/RxJava/javadoc/io/reactivex/Observable.html#repeat-long-).
  */
-fun <T> Observable<T>.repeat(count: Int = -1): Observable<T> =
+fun <T> Observable<T>.repeat(times: Long = Long.MAX_VALUE): Observable<T> =
     observable { emitter ->
         val observer =
             object : ObservableObserver<T>, ValueCallback<T> by emitter, ErrorCallback by emitter {
-                private val counter: AtomicInt? = if (count >= 0) AtomicInt(count) else null
+                private val counter: AtomicLong? = if (times >= 0) AtomicLong(times) else null
 
                 // Prevents recursive subscriptions
                 private val serializer =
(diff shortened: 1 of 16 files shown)


reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/Repeat.kt
```diff
@@ -4,18 +4,19 @@ import com.badoo.reaktive.base.ErrorCallback
...
 import com.badoo.reaktive.utils.serializer.serializer
 
 /**
- * Returns an [Observable] that automatically resubscribes to this [Observable] at most [count] times.
+ * Returns an [Observable] that automatically resubscribes to this [Observable] at most [times] times.
  *
  * Please refer to the corresponding RxJava [document](http://reactivex.io/RxJava/javadoc/io/reactivex/Observable.html#repeat-long-).
  */
-fun <T> Observable<T>.repeat(count: Int = -1): Observable<T> =
+fun <T> Observable<T>.repeat(times: Long = Long.MAX_VALUE): Observable<T> =
```
arkivanov, line 15: Let's add the following check before returning the observable:
```kotlin
require(times >= 0L) { "Number of times must not be negative" }
```
RxJava does the same, so we need consistency here as well.

reaktive/src/commonTest/kotlin/com/badoo/reaktive/completable/RetryTest.kt
```diff
@@ -79,7 +79,7 @@ class RetryTest : CompletableToCompletableTests by CompletableToCompletableTests
 
     @Test
     fun predicate_receives_valid_counter_WHEN_upstream_produces_error() {
-        val timeRef = AtomicInt()
+        val timeRef = AtomicLong(0)
```
arkivanov, line 82: What kind of warning are you getting here?

reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/RepeatTest.kt
```diff
@@ -44,7 +44,7 @@ class RepeatTest : ObservableToObservableTests by ObservableToObservableTestsImp
     @Test
     fun emits_all_values_of_first_iteration_WHEN_count_is_negative() {
         val upstream = TestObservable<Int?>()
-        val observer = upstream.repeat(count = -1).test()
+        val observer = upstream.repeat(times = -1).test()
```
arkivanov, line 47: This test is now invalid, because negative values are now incorrect. It should be `Long.MAX_VALUE`. Same for other places where `-1` is passed.

arkivanov (changes requested):

arkivanov:
Also could you please run apiDump Gradle task and commit the updated api files? Otherwise build will fail. Don't forget to update the branch first, because of #647.

arkivanov:
Also I noticed an inconsistent (with RxJava) behaviour in the `retry` operator, which is there from the very beginning. The `retry` operator should return `empty` `Observable`, if `times` is 0. It should subscribe only once to the upstream if `times` is 1, the upstream should not repeat. If `times` is 2, the upstream should be repeated only once. And so on. Currently there is no way to get an empty `Observable`, and the subscription always happens at least one.

I added a separate item in the 2.0.0 fixes list for this.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/Repeat.kt
emartynov: Done

reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/RepeatTest.kt
emartynov: Fixed

reaktive/src/commonTest/kotlin/com/badoo/reaktive/completable/RetryTest.kt
emartynov: Fixed

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/Repeat.kt
```diff
@@ -3,19 +3,21 @@ package com.badoo.reaktive.observable
...
-fun <T> Observable<T>.repeat(count: Int = -1): Observable<T> =
-    observable { emitter ->
+fun <T> Observable<T>.repeat(times: Long = Long.MAX_VALUE): Observable<T> {
+    require(times >= 0L) { "Number of times must not be negative" }
+
+    return observable { emitter ->
         val observer =
             object : ObservableObserver<T>, ValueCallback<T> by emitter, ErrorCallback by emitter {
-                private val counter: AtomicInt? = if (count >= 0) AtomicInt(count) else null
+                private val counter: AtomicLong? = if (times >= 0) AtomicLong(times) else null
```
arkivanov, line 20: The condition `times >= 0` is always `true`. Looks like it should now be `if (times < Long.MAX_VALUE) AtomicLong(times) else null`?

arkivanov (changes requested):

arkivanov:
Also there conflicts with the base breach, would it be possible for you to resolve them?

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/Repeat.kt
emartynov: Fixed

emartynov:
Resolved also conflicts

arkivanov (approved):

arkivanov:
Thanks!

PR #662, opened 2022-05-01
Updated Kotlin to 1.6.21 with HMPP enabled

arkivanov:
This PR updates Kotlin to 1.6.21 and coroutines to 1.6.1. Since Kotlin 1.6.20, HMPP is enabled by default. So this PR is a follow-up to #638.

[KT-48811](https://youtrack.jetbrains.com/issue/KT-48811) is marked as fixed, however it's still relevant and is now being tracked in [KT-52193](https://youtrack.jetbrains.com/issue/KT-52193). Because of this, I had to supply default values to atomics in shared source sets, even though they have default values specified in `expect` common.

### DIFF
--- a/reaktive/src/darwinCommonMain/kotlin/com/badoo/reaktive/scheduler/MainScheduler.kt
+++ b/reaktive/src/darwinCommonMain/kotlin/com/badoo/reaktive/scheduler/MainScheduler.kt
@@ -59,7 +59,7 @@ internal class MainScheduler : Scheduler {
         }
 
         override fun cancel() {
-            operations.clear()
+            operations.clear(dispose = true)
         }
 
         override val isDisposed: Boolean
--- a/reaktive/src/jsMain/kotlin/com/badoo/reaktive/utils/SynchronizedCompat.kt
+++ b/reaktive/src/jsMain/kotlin/com/badoo/reaktive/utils/SynchronizedCompat.kt
@@ -1,0 +1,3 @@
+package com.badoo.reaktive.utils
+
+internal actual inline fun <T> synchronizedCompat(lock: Any, block: () -> T): T = block()
--- a/build.gradle
+++ b/build.gradle
@@ -10,6 +10,7 @@ plugins {
 allprojects {
     repositories {
         google()
+        mavenCentral()
         jcenter()
     }
 }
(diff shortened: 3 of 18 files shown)


arkivanov:
@CherryPerry Could you please run builds? I want to check if it compiles on CI. Thanks!

reaktive/src/jsMain/kotlin/com/badoo/reaktive/utils/SynchronizedCompat.kt
```diff
@@ -0,0 +1,3 @@
+package com.badoo.reaktive.utils
+
+internal actual inline fun <T> synchronizedCompat(lock: Any, block: () -> T): T = block()
```
CherryPerry, line 3: Should not `synchronized` be no-op in JS by default?
Or it was removed?

arkivanov: It was deprecated for a long time, and now doesn't compile in a shared source set. However it's still present in JS, and deprecated. Not sure what is the actual reason, but better to not use deprecated things anyway.

arkivanov:
@CherryPerry This PR is ready for review, thanks!

reaktive/src/darwinCommonMain/kotlin/com/badoo/reaktive/scheduler/MainScheduler.kt
```diff
@@ -59,7 +59,7 @@ internal class MainScheduler : Scheduler {
         }
 
         override fun cancel() {
-            operations.clear()
+            operations.clear(dispose = true)
```
CherryPerry, line 62: `true` is a default value.

CherryPerry (approved):

reaktive/src/darwinCommonMain/kotlin/com/badoo/reaktive/scheduler/MainScheduler.kt
arkivanov: Yep, as I mentioned in the description, there is a bug and we must supply the value in shared source sets.

arkivanov: Same for atomics

Issue #665, opened 2022-06-09
Update Kotlin to 1.7.0


arkivanov:
What's new: https://kotlinlang.org/docs/whatsnew17.html

PR #666, opened 2022-06-10
Updated Kotlin to 1.7.0, Gradle to 7.4.2, AGP to 7.2.0, binary-compatibility-plugin to 0.10.0

### LINKED ISSUE #663: Possible problems during upgrade to Kotlin 1.7.0
Hey,

I've discovered the issue of upgrading the Reaktive project on the 1.7.0 Kotlin version (on RC). The IDE import and project build may fail with errors about binary incompatibility. For example: 
```
> Task :gradleConfiguration:compileKotlin FAILED
e: /.gradle/caches/modules-2/files-2.1/org.jetbrains.kotlin/kotlin-compiler-embeddable/1.7.0-RC-211/f5185a225bcd8d4d47023bae2b76b687aba0d372/kotlin-compiler-embeddable-1.7.0-RC-211.jar!/META-INF/backend-common.kotlin_module: Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.7.1, expected version is 1.5.1.
```

The problem here is the `org.jetbrains.kotlin:kotlin-compiler-embeddable` dependency inside `gradleConfiguration`. It includes a `1.7.0` version of the artifact to the project. But it conflicts with the version required by the Kotlin 1.5.31 bundled to Gradle wrapper that also depends on `kotlin-compiler-embeddable`. The outdated Kotlin was involved in the process since the module for composite build uses `kotlin-dsl` plugin for configuration. 
All that leads to the incompatible difference in metadata in more than 1 minor version (1.5.1 vs 1.7.1).

Suggested workaround:
* Upgrade Gradle wrapper to 7.5 (currently in RC but maybe after the release of 1.7.0 will be available in Stable). It will include 1.6.21 Kotlin. 
* Keep `kotlin-compiler-embeddable` dependency with the 1.6.20 version.

No action is required right now, just to be aware.
Thanks for the library!

arkivanov:
Changes:

- Updated Gradle to `7.4.2`
- Updated Kotlin to `1.7.0`
- Removed `kotlin-compiler-embeddable` dependency from `gradleConfiguration` module (looks like it is unused, also see #663)
- Updated `binary-compatibility-validator` to `0.10.0`
- Updated AGP to `7.2.0`
- Fixed some warnings in the code

Closes #663, #665


arkivanov:
@CherryPerry Could you please review?

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.2-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.4.2-bin.zip
```
CherryPerry, line 3: Can you please use `all` here.
Also can you upgrade version with `./gradlew wrapper --gradle-version=7.4.2 --distribution-type=all` (in case if you did not) so other related files can also be updated.

arkivanov: Done

arkivanov:
@CherryPerry I have addressed the comment above and resolved merge conflicts

CherryPerry (approved):

PR #674, opened 2022-07-22
Added `using` operators

### LINKED ISSUE #673: Add `using` operators
Docs: https://reactivex.io/documentation/operators/using.html

arkivanov:
Closes #673 

cc: @CherryPerry

### DIFF
--- a/reaktive/src/commonTest/kotlin/com/badoo/reaktive/completable/UsingTest.kt
+++ b/reaktive/src/commonTest/kotlin/com/badoo/reaktive/completable/UsingTest.kt
@@ -1,0 +1,167 @@
+package com.badoo.reaktive.completable
+
+import com.badoo.reaktive.disposable.Disposable
+import com.badoo.reaktive.test.completable.DefaultCompletableObserver
+import com.badoo.reaktive.test.completable.TestCompletable
+import com.badoo.reaktive.test.completable.test
+import com.badoo.reaktive.utils.SharedList
+import com.badoo.reaktive.utils.atomic.AtomicBoolean
+import com.badoo.reaktive.utils.atomic.getValue
+import com.badoo.reaktive.utils.atomic.setValue
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+
+class UsingTest :
+    CompletableToCompletableTests by CompletableToCompletableTestsImpl(
+        transform = { completableUsing(resourceSupplier = {}, resourceCleanup = {}, sourceSupplier = { this }) }
+    ) {
+
+    private val disposables = Disposables()
+
+    @Test
+    fun acquires_new_resource_each_time_WHEN_eager_is_true_and_subscribed_multiple_times() {
+        val downstream = completableUsing(eager = true)
+
+        repeat(3) { downstream.test() }
+
+        assertEquals(3, disposables.size)
+    }
+
+    @Test
+    fun acquires_new_resource_each_time_WHEN_eager_is_false_and_subscribed_multiple_times() {
+        val downstream = completableUsing(eager = false)
+
+        repeat(3) { downstream.test() }
+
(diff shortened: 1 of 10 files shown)


reaktive/src/commonTest/kotlin/com/badoo/reaktive/completable/UsingTest.kt
```diff
@@ -0,0 +1,256 @@
...
+            }
+        )
+
+        upstream.onComplete()
+
+        assertFalse(isResourceDisposedBeforeInnerCompletable)
+    }
+
+    @Test
+    fun resource_is_not_disposed_WHEN_eager_is_false_and_downstream_received_onError() {
```
CherryPerry, line 225: Weird test name, it is hard to understand what you are testing here. `using` by design should invoke `resourceCleanup` but test name says it should not.

```kotlin
@Test
fun resource_is_not_disposed _BEFORE_DOWNSTREAM_ WHEN_eager_is_false_and_downstream_received_onError()
```

reaktive/src/commonTest/kotlin/com/badoo/reaktive/completable/UsingTest.kt
```diff
@@ -0,0 +1,256 @@
...
+import com.badoo.reaktive.utils.atomic.AtomicList
+import com.badoo.reaktive.utils.atomic.getValue
+import com.badoo.reaktive.utils.atomic.plusAssign
+import com.badoo.reaktive.utils.atomic.setValue
+import kotlin.test.Test
+import kotlin.test.assertContentEquals
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+
+class UsingTest :
```
CherryPerry, line 18: You can create `class DisposablesFactoryStub: () -> Disposable` that will store all created objects that you can verify later. By doing this we can verify that we invoke `resourceSupplier` only once (or amount of times we needed). Also we can verify that we do not create resource when it is not required.

```kotlin
    @Test
    fun acquires_new_resource_each_time_WHEN_eager_is_true_and_subscribed_multiple_times() {
        val factory = DisposablesFactoryStub()
        val acquiredResources = AtomicList<Disposable>(emptyList())

        val downstream =
            completableUsing(resourceSupplier = factory, eager = true) {
                acquiredResources += it
                TestCompletable()
            }

        repeat(3) { downstream.test() }

        assertContentEquals(factory.created, acquiredResources.value)
    }
```

reaktive/src/commonTest/kotlin/com/badoo/reaktive/completable/UsingTest.kt
arkivanov: There are two modes: `eager=true` and `eager=false`. The former should dispose the resource before terminating the downstream, the latter should dispose the resource after terminating the downstream. So, when the downstream receives either `onComplete` or `onError`, the resource should already be disposed if `eager=true` and should be yet disposed if `eager=false`.

This test says - resource is not disposed WHEN downstream received `onError`. Perhaps a better name would be "resource_is_not_yet_disposed_WHEN_eager_is_false_and_downstream_received_onError`.

CherryPerry: What do you think about `resource_is_not_disposed_BEFORE_DOWNSTREAM_`?

arkivanov: I have updated all tests, now it says `disposes_resource_after_downstream_signalled_onError_WHEN_eager_false_and_upstream_produced_error`.

reaktive/src/commonTest/kotlin/com/badoo/reaktive/completable/UsingTest.kt
arkivanov: Fixed.

CherryPerry (approved):

PR #676, opened 2022-07-27
Update action versions

CherryPerry:
@arkivanov

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,9 +7,6 @@ on:
       - mt-coroutines
   pull_request:
 
-env:
-  GRADLE_OPTS: -Dorg.gradle.daemon=false -Dorg.gradle.workers.max=2 -Dkotlin.compiler.execution.strategy=in-process
-
 jobs:
   build:
     strategy:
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -19,33 +16,28 @@ jobs:
...
-        uses: actions/setup-java@v1
+        uses: actions/setup-java@v3
         with:
-          java-version: 11
+          distribution: 'zulu'
+          java-version: '11'
       - name: Install dependencies
         if: matrix.os == 'ubuntu-latest'
         run: sudo apt-get update && sudo apt-get install libcurl4-openssl-dev libgtk-3-dev nodejs chromium-browser
-      - name: Set up cache
```
arkivanov, line 30: Why did you remove this?

.github/workflows/release-nmtc.yml
```diff
@@ -15,12 +12,12 @@ jobs:
     steps:
       - id: create
         name: Create staging repository
-        uses: nexus-actions/create-nexus-staging-repo@v1
+        uses: nexus-actions/create-nexus-staging-repo@67d9c1133c2fdcb777e20c040595073e20311225
```
arkivanov, line 15: Why not a stable version?

.github/workflows/build.yml
CherryPerry: `gradle/gradle-build-action` manages caches keys by itself, better to rely on Gradle team rather than managing it ourselves.

.github/workflows/release-nmtc.yml
CherryPerry: I am ready to rely on well-known actions maintained by teams, but this one unfortunately is not and we are passing very critical information there.

Also it is [advice](https://github.com/nexus-actions/create-nexus-staging-repo) from the author:

```       
# replace '@main' with the latest commit sha, see https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions
```

.github/workflows/build.yml
arkivanov: I thought that GitHub Actions drops everything including cache folders by default. But if you believe it's ok then lgtm.

CherryPerry: [gradle/gradle-build-action](https://github.com/gradle/gradle-build-action#caching)

.github/workflows/release-nmtc.yml
arkivanov: TIL

.github/workflows/build.yml
arkivanov: TIL

arkivanov (approved):

PR #687, opened 2022-09-29
implemented coroutines-swing lib

Tolegen99:


arkivanov:
Thanks for the PR! Could you explain why is this necessary? There are few main dispatcher implementations available, like Swing, JavaFx, Android. Why Swing in commonMain?

PR #689, opened 2022-12-07
Updated Kotlin to 1.7.21

arkivanov:
Since Kotlin 1.7.20 the new K/N memory model is enabled by default. Since Reaktive still supports the legacy model, some of our tests fail if freezing is disabled. I enabled the legacy model for now, while Reaktive supports it (perhaps until v2.0). This should not affect clients' experience.


CherryPerry (approved):

arkivanov:
@CherryPerry Could you please re-run the build? The Gradle daemon disappeared unexpectedly.

PR #691, opened 2023-01-16
Updated Kotlin to 1.8.0

### LINKED ISSUE #690: Update Kotlin to 1.8.0
What's new: https://kotlinlang.org/docs/whatsnew18.html

arkivanov:
Kotlin 1.8.0
AGP 7.3.1
Gradle 7.5

Closes #690

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.4.2-all.zip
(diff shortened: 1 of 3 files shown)


CherryPerry:
It is a little bit late, but we can do a minor release first with Kotlin 1.7 just in case. But looks like nobody asked for a separate release, so I assume they 1.6 (released) and 1.7 (unreleased) are compatible between each other. WDYT?

CherryPerry (dismissed):

arkivanov:
I think a library compiled with Kotlin 1.6 can not be used in a project with Kotlin 1.8, because of metadata compatibility. It might make sense to release a new version before merging this PR.

I will amend the commit a bit later, no problem.

arkivanov:
Amended the commit.

CherryPerry (approved):

PR #693, opened 2023-01-18
Hardcode Xcode version because Kotlin expects particular emulators

CherryPerry:
Looks like it is better to match Xcode version with particular Kotlin version.

```
java.lang.IllegalStateException: command '/usr/bin/xcrun' exited with errors (exit code: 148)
Invalid device: Apple Watch Series 5 - 44mm
```

Issue: https://youtrack.jetbrains.com/issue/KT-54814


arkivanov:
This issue should be fixed in Kotlin 1.8.0, for now we can use macos-11, see https://github.com/arkivanov/Decompose/pull/301 for an example

CherryPerry:
Will keep the current approach and replace after 1.8 is merged.

PR #699, opened 2023-02-22
Removed samples

arkivanov:
The samples are super simple and so not very useful. Also, there are plenty of examples on GitHub (updated the Readme).

Another point is that it's difficult to compile the Linux and iOS samples, because additional libraries must be installed in the system (e.g. curl and Cocoapods on macOS).

On the other hand, CI times will much faster now.


CherryPerry:
Not sure it is a good idea. Samples works not only as sample, but also as verification project that integration into real apps actually works. Which is important especially for Linux and Darwin targets.

arkivanov:
Good point about playground. Maybe instead we should update samples, so that there is no cinterop and network access (Cocoapods). Otherwise it's extremely painful to build the whole project on macOS.

arkivanov:
Closed in favour of #703

PR #714, opened 2023-03-02
[2.0] Removed SharedQueue and replaced all usages with ArrayDeque

arkivanov:


reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/ConcatMap.kt
```diff
@@ -32,7 +31,7 @@ private class ConcatMapObserver<in T, in R>(
 
     private val actor = serializer(::processEvent)
     private val innerObserver = InnerObserver(callbacks, actor).addTo(this)
-    private val queue = SharedQueue<T>()
```
CherryPerry, line 35: Remove `SharedQueue` implementation too?

arkivanov: Oh, thanks. Missed that. Will do.

arkivanov: Removed.

CherryPerry (approved):

PR #716, opened 2023-03-03
[2.0] Removed LinkedFreezableQueue and optimized BufferedExecutor

arkivanov:


CherryPerry (approved):

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/ObserveOn.kt
```diff
@@ -19,7 +19,6 @@ fun <T> Observable<T>.observeOn(scheduler: Scheduler): Observable<T> =
         val executor = scheduler.newExecutor()
         disposables += executor
         val bufferedExecutor = BufferedExecutor(executor, emitter::onNext)
-        disposables += bufferedExecutor
```
CherryPerry, line 22: Should not we `cancel` the executor? Emissions will be omitted because emitter is disposed, but we will still try to execute it for each task in  `BufferedExecutor` queue.

CherryPerry: `executor` is disposed, so it will work fine.

arkivanov: It appears that disposing BufferedExecutor now has no effect. Before all the changes, on native `BufferedExecutor#dispose` was setting `isDisposed` flag to true and disposing the `Lock`, on JVM only the flag was changed. The flag is never read, so there is no need to `dispose` at all. The behaviour seems unchanged.

Also, cancelling `BufferedExecutor` seems unnecessary, because it checks `Executor#isDisposed` on each iteration.

Does it look good?

PR #718, opened 2023-03-09
[2.0] Removed ArrayQueue and optimized Serializer

arkivanov:
- Removed `ArrayQueue` and replaced all usages with `ArrayDeque` from stdlib
- Merged native and non-native implementations of `Serializer` into one and move to common
- Optimized `Serializer` a bit, so it uses only one atomic instead of two
- Added concurrency tests for `Serializer`


reaktive/src/commonMain/kotlin/com/badoo/reaktive/utils/serializer/DefaultSerializer.kt
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+internal abstract class DefaultSerializer<T> : AbstractSerializer<T>() {
+    private var queue: ArrayDeque<T>? = null
+
+    override fun addLast(value: T) {
+        if (queue == null) {
+            queue = ArrayDeque()
+        }
+        queue!!.add(value)
```
CherryPerry, line 18: > queue!!

Use same approach as in the other place please
`val q = queue ?: ArrayDeque<T>().also { queue = it }`

reaktive/src/commonMain/kotlin/com/badoo/reaktive/utils/serializer/PrioritySerializer.kt
```diff
@@ -0,0 +1,37 @@
...
+    private val comparator: Comparator<in T>,
+) : AbstractSerializer<T>() {
+
+    private var queue: PriorityQueue<T>? = null
+
+    override fun addLast(value: T) {
+        if (queue == null) {
+            queue = PriorityQueue(comparator)
+        }
+        queue!!.offer(value)
```
CherryPerry, line 24: Same.

reaktive/src/jvmTest/kotlin/com/badoo/reaktive/utils/serializer/DefaultSerializerThreadingTest.kt
```diff
@@ -0,0 +1,93 @@
...
+        repeat(THREAD_COUNT) { threadIndex ->
+            doInBackground {
+                startLatch.countDown()
+                startLatch.await()
+
+                repeat(ITERATION_COUNT) { iterationIndex ->
+                    block(threadIndex, iterationIndex)
+                }
+
+                finishLatch.countDown()
```
CherryPerry, line 82: `finishLatch.countDown()` should be in `finally` block to not dead lock test execution in case if exception is thrown in `block`.

Can you also check what happens in case of exception in `block`? Test fails? Test suite fails? Exception ignored (I assume this variant). Most likely you need to capture the exception and rethrow it?

CherryPerry (changes requested):

CherryPerry (approved):

PR #719, opened 2023-03-09
Build cache tweak

CherryPerry:


.github/workflows/build.yml
```diff
@@ -36,13 +37,17 @@ jobs:
         if: matrix.os == 'ubuntu-latest'
         uses: gradle/gradle-build-action@v2
         with:
+          gradle-home-cache-cleanup: true
+          cache-read-only: ${{ github.ref != 'refs/heads/master' && github.ref != 'refs/heads/version-2.0' }}
```
arkivanov, line 41: Maybe also add mt-coroutines here?

CherryPerry: Will be dropped in 2.0 anyway, highly doubt that there will be any changes there.

PR #721, opened 2023-03-11
Use version catalog

CherryPerry:
@arkivanov Fixed memory leak in the corresponding commit, `Resources` was initialized twice, but destroyed once.


arkivanov:
> Fix memory leak

This doesn't look like an actual memory leak, since `Cleaner` should be triggered by GC anyway after there are no references to `Condition`. But thanks for fixing!

PR #726, opened 2023-03-15
[2.0] Removed AtomicList typealias

arkivanov:


reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/WithLatestFrom.kt
```diff
@@ -26,7 +24,8 @@ fun <T, U, R> Observable<T>.withLatestFrom(
         emitter.setDisposable(disposables)
 
         val otherSources = others.toList()
-        val otherValues = atomicList<Any?>(List(otherSources.size) { Uninitialized })
+        val otherValues = HashMap<Int, U>()
```
CherryPerry, line 27: Maybe 2 arrays instead (lastValue + isInitialized)? Should be both faster and more memory efficient.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/Zip.kt
```diff
@@ -30,7 +27,7 @@ fun <T, R> Iterable<Observable<T>>.zip(mapper: (List<T>) -> R): Observable<R> =
         emitter.setDisposable(disposables)
 
         val values = List<ArrayList<T>>(sources.size) { ArrayList() }
-        val completed = AtomicList(List(sources.size) { false })
+        val completed = MutableList(sources.size) { false }
```
CherryPerry, line 30: Also probably makes sense to revisit operators that works with ArrayLists of fixed size and replace with arrays. Almost free win at least on boxing-unboxing.

But up to you.

CherryPerry (dismissed):

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/WithLatestFrom.kt
arkivanov: This would require iterating over the `isInitialized` array every time in the upstream's `onNext` method? Or we would need to also memoize it with a flag. WDYT?

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/Zip.kt
arkivanov: Yes, this makes sense. Will update this one.

reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/WithLatestFrom.kt
arkivanov: Fixed, please review

CherryPerry (approved):

PR #729, opened 2023-04-17
Undesirable disconnect fix with replay, refCount, and firstOrError

bjsvedin:
Modified the refCount function to account for the subscribeCount before the actual subscribe. This resolved an issue when you are using both replay, and firstOrError.

I have attempted to create a Unit test to replicate the issue but I was unsuccessful, I am not familiar enough with the testing environment to recreate it in Unit Tests. I  have left the test in comments however.

To replicate the issue: 
 - create an observable
 - apply replay(1).refcount() and store separately
 - subscribe once to the second observable which causes a connection
 - send a value through the upstream
 - THEN subscribe again to the second observable using firstOrError()

This results in the connectable being disconnected. The reason being is with replay and a value stored in the buffer, the single subscribes, receives the onNext and immediately completes before the refCounts subscribeCount increase. So when the single completes and disposes, the subscribeCount is decremented to 0 and it disposes the connectable even though there is still a subscriber active.

### DIFF
--- a/reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/RefCountTest.kt
+++ b/reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/RefCountTest.kt
@@ -1,11 +1,13 @@
 package com.badoo.reaktive.observable
 
 import com.badoo.reaktive.disposable.Disposable
+import com.badoo.reaktive.subject.publish.PublishSubject
 import com.badoo.reaktive.test.base.assertError
 import com.badoo.reaktive.test.observable.TestObservableObserver
 import com.badoo.reaktive.test.observable.assertComplete
 import com.badoo.reaktive.test.observable.assertValues
 import com.badoo.reaktive.test.observable.test
+import com.badoo.reaktive.test.single.test
 import com.badoo.reaktive.utils.SharedList
 import com.badoo.reaktive.utils.atomic.AtomicBoolean
 import com.badoo.reaktive.utils.atomic.AtomicInt
@@ -111,6 +113,18 @@ class RefCountTest {
         assertFalse(disposable.isDisposed)
     }
 
+    // @Test
+    // fun does_not_disconnect_from_upstream_WHEN_subscriberCount_is_1_and_first_or_error_with_replay() {
+    //     val disposable = Disposable()
+    //     val upstream = PublishSubject<Int>()
+    //     val refCount = upstream.replay(1).refCount(subscriberCount = 1)
+    //     refCount.test()
+    //     upstream.onNext(1)
+    //     refCount.firstOrError().test()
+    //
+    //     assertFalse(disposable.isDisposed)
+    // }
+
     @Test
     fun subscription_to_upstream_happens_before_connection_to_upstream() {
         val events = SharedList<String>()
@@ -214,4 +228,5 @@ class RefCountTest {
                 subscribe.invoke(observer)
             }
         }
+
 }
--- a/reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/RefCountTest.kt
+++ b/reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/RefCountTest.kt
@@ -245,5 +245,4 @@ class RefCountTest {
                 subscribe.invoke(observer)
             }
         }
-
 }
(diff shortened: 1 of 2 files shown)


arkivanov:
Thanks for the PR! This fix looks good.

Could you please add the following test?

```kotlin
    @Test
    fun does_not_disconnect_WHEN_connected_and_new_observer_disposed_synchronously_on_emission() {
        val disposable = Disposable()

        val upstream =
            testUpstream(
                connect = { onConnect -> onConnect?.invoke(disposable) },
                subscribe = { observer ->
                    observer.onSubscribe(Disposable())
                    observer.onNext(1)
                }
            )

        val refCount = upstream.refCount(subscriberCount = 1)

        refCount.test() // First subscription

        refCount.subscribe(
            object : DefaultObservableObserver<Int?> {
                private var disposableRef: Disposable? by AtomicReference(null)

                    disposableRef = disposable
                }

                override fun onNext(value: Int?) {
                    requireNotNull(disposableRef).dispose()
                }
            }
        )

        assertFalse(disposable.isDisposed)
    }
```

arkivanov (changes requested):
LGTM, but let's add the test mentioned in the conversation.

reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/RefCountTest.kt
```diff
@@ -111,6 +113,18 @@ class RefCountTest {
...
+    //     val disposable = Disposable()
+    //     val upstream = PublishSubject<Int>()
+    //     val refCount = upstream.replay(1).refCount(subscriberCount = 1)
+    //     refCount.test()
+    //     upstream.onNext(1)
+    //     refCount.firstOrError().test()
+    //
+    //     assertFalse(disposable.isDisposed)
+    // }
+
```
arkivanov, line 127:
```suggestion
    @Test
    fun does_not_disconnect_WHEN_connected_and_new_observer_disposed_synchronously_on_emission() {
        val disposable = Disposable()

        val upstream =
            testUpstream(
                connect = { onConnect -> onConnect?.invoke(disposable) },
                subscribe = { observer ->
                    observer.onSubscribe(Disposable())
                    observer.onNext(1)
                }
            )

        val refCount = upstream.refCount(subscriberCount = 1)

        refCount.test() // First subscription

        refCount.subscribe(
            object : DefaultObservableObserver<Int?> {
                private var disposableRef: Disposable? by AtomicReference(null)

                override fun onSubscribe(disposable: Disposable) {
                    disposableRef = disposable
                }

                override fun onNext(value: Int?) {
                    requireNotNull(disposableRef).dispose()
                }
            }
        )

        assertFalse(disposable.isDisposed)
    }
```

reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/RefCountTest.kt
```diff
@@ -214,4 +228,5 @@ class RefCountTest {
                 subscribe.invoke(observer)
             }
         }
+
```
arkivanov, line 231:
```suggestion
```

arkivanov (changes requested):

arkivanov (dismissed):

arkivanov (approved):

reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/RefCountTest.kt
bjsvedin: suggestion implemented.

reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/RefCountTest.kt
bjsvedin: Suggestion implemented, though I couldn't use AtomicReference as a delgate so I used it directly.

bjsvedin:
I have implemented the suggestions provided.

reaktive/src/commonTest/kotlin/com/badoo/reaktive/observable/RefCountTest.kt
arkivanov: This is fine, thanks!

CherryPerry (approved):

PR #731, opened 2023-04-21
[2.0] Make atomics extend JVM atomics

arkivanov:
Currently our atomics on JVM create Java atomics privately. This leads to additional allocations. Since our atomics are now internal, we can make them extend Java atomics and so save allocations. This makes some existing extension functions to clash with the methods from Java atomics, like `getAndUpdate`, `updateAndGet`, etc. So I had to rename those methods.

Also used normal `AtomicReference` instead of `FreezableAtomicReference` in K/N, because the latter is deprecated and not required with the new memory model.


CherryPerry (commented):
In general I am not against the proposed changes, but it feels a little bit weird in a couple of things:
1. We are allocating 1 more object mostly only during subscription. So it is kinda negligible, comparing to having a different set of API to use.
2. You have mentioned kotlinx-atomicfu in the previous PR, can we use it instead? It already has the same API as we have (update/getAndUpdate).

CherryPerry (approved):

PR #736, opened 2023-04-28
[2.0] Optimize expect/actual boilerplate

arkivanov:
Some of the expect/actual implementations are no longer required, can be just implemented in common. Also made some additional performance optimizations.


reaktive/src/commonMain/kotlin/com/badoo/reaktive/disposable/SerialDisposable.kt
```diff
@@ -14,6 +14,14 @@ expect open class SerialDisposable() : Disposable {
      */
     override fun dispose()
 
+    /**
+     * Atomically removes any existing [Disposable] and disposes this [SerialDisposable].
+     * Does not dispose the removed [Disposable].
+     *
+     * @return the removed [Disposable] if any.
+     */
+    internal fun clearAndDispose(): Disposable?
```
CherryPerry, line 23: Can be implemented as an extension function to not pollute public class with internal functions.
```kotlin
val d = replace(null)
dispose()
return d
```

arkivanov: The idea is to perform the operation under one `synchronized` block.

CherryPerry (approved):

PR #744, opened 2023-05-16
Make buffer limit arguments of type int instead of long

arkivanov:
Aligned with RxJava.

As part of #620.

### DIFF
--- a/reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/Buffer.kt
+++ b/reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/Buffer.kt
@@ -26,10 +26,10 @@ fun <T> Observable<T>.buffer(
  */
 fun <T> Observable<T>.buffer(
     boundaries: Observable<*>,
-    limit: Long = Long.MAX_VALUE,
+    limit: Int = Int.MAX_VALUE,
     restartOnLimit: Boolean = false
 ): Observable<List<T>> =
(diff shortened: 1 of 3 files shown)


reaktive/src/commonMain/kotlin/com/badoo/reaktive/observable/Buffer.kt
```diff
@@ -57,8 +57,8 @@ fun <T> Observable<T>.buffer(
 fun <T, S> Observable<T>.buffer(
     opening: Observable<S>,
     closing: (S) -> Completable,
-    limit: Long = Long.MAX_VALUE,
+    limit: Int = Int.MAX_VALUE,
     restartOnLimit: Boolean = false
 ): Observable<List<T>> =
-    window(opening = opening, closing = closing, limit = limit, restartOnLimit = restartOnLimit)
+    window(opening = opening, closing = closing, limit = limit.toLong(), restartOnLimit = restartOnLimit)
```
CherryPerry, line 63: Are you going to update `window` too?

arkivanov: Window looks fine with long, it doesn't have any buffers, just `Long` counters. RxJava also uses `Long` for windows - http://reactivex.io/RxJava/javadoc/io/reactivex/Observable.html#window-long-java.util.concurrent.TimeUnit-io.reactivex.Scheduler-long-boolean-

CherryPerry (approved):

PR #746, opened 2023-06-04
[2.0] Move Scheduler interop with coroutines to commonMain

arkivanov:
Also refactored  `CoroutineContextScheduler` so it doesn't use deprecated channels anymore.


arkivanov:
The implementation is a bit incorrect, I will fix and open another PR.

PR #748, opened 2023-06-05
[2.0] Move Scheduler interop with coroutines to commonMain

arkivanov:
Also refactored  `CoroutineContextScheduler` so it doesn't use any deprecated or legacy APIs. And also fixed a bug discovered during refactoring, when new tasks are not being executed once the `Executor` is cancelled (not disposed).


coroutines-interop/src/commonMain/kotlin/com/badoo/reaktive/coroutinesinterop/CoroutineContextScheduler.kt
```diff
@@ -37,71 +34,53 @@ internal class CoroutineContextScheduler(
...
 
         override fun submit(delay: Duration, period: Duration, task: () -> Unit) {
-            channelRef.value?.channel?.apply {
-                trySend(
-                    Task(
-                        startTime = clock.uptime + delay.coerceAtLeastZero(),
-                        period = period.coerceAtLeastZero(),
-                        task = task
-                    )
+            channelRef.value?.channel?.trySend(
```
CherryPerry, line 68: What if `trySend` will fail? I can imagine the following case:
1. Create this class with `ChannelHolder` on `thread-1`
2. It will enqueue channel collection task to be executed on `thread-2` (single-threaded coroutine context) via `scope.launch`
3. `thread-3` will trigger `submit`
4. `trySend` will fail, collector is not ready yet
5. Collector is ready on `thread-3`
6. Task is lost

You can try to cover this case in tests with `StandardTestDispatcher`.

CherryPerry: Actually never mind. `trySend` always successes `Channel.UNLIMITED` capacity.
Hard to understand it from code without comments.

CherryPerry (approved):

coroutines-interop/src/commonMain/kotlin/com/badoo/reaktive/coroutinesinterop/CoroutineContextScheduler.kt
arkivanov: Yep, `trySend` can only fail only on buffer overflow, which shouldn't happen with `UNLIMITED` capacity. I could add a comment if you like. Let me know.

PR #774, opened 2024-02-24
Fixed a race condition in refCount

arkivanov:
Fixed a race condition in `refCount` when the second connection is performed concurrently while the first connection is still being disconnected.
Also fixed a bug when the upstream is not connected when disposed synchronously.


CherryPerry (approved):

arkivanov:
Thanks for pointing that out! At first glance it looks not applicable, but I will check further later.

Issue #772, opened 2024-02-17
Use macos-14 runner for better build speed


arkivanov:
The new Apple Silicon `macos-14` runner is now available for free, and it should be much faster.

PR #775, opened 2024-03-11
Workflows update

CherryPerry:


arkivanov:
Maybe also resolve #772?

PR #777, opened 2024-06-03
Updated Kotlin to 2.0.0, coroutines to 1.8.1, Detekt to 1.23.6, AGP to 8.2.0

### LINKED ISSUE #776: Update Kotlin to 2.0.0

arkivanov:
Also added logic to avoid exporting the sample iOS framework if not building for Xcode. This makes CI builds much faster, as it doesn't link all architectures.

Closes #776.

### DIFF
--- a/sample-android-app/build.gradle
+++ b/sample-android-app/build.gradle
@@ -5,12 +5,12 @@ plugins {
 
 android {
     namespace "com.badoo.reaktive.sample.android"
-    compileSdk 31
+    compileSdk 33
 
     defaultConfig {
         applicationId 'com.badoo.reaktive.sample.android'
         minSdkVersion 15
-        targetSdkVersion 31
+        targetSdkVersion 33
         versionCode 1
         versionName '1.0'
     }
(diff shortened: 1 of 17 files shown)


sample-android-app/build.gradle
```diff
@@ -5,12 +5,12 @@ plugins {
 
 android {
     namespace "com.badoo.reaktive.sample.android"
-    compileSdk 31
+    compileSdk 33
```
CherryPerry, line 8: 34 maybe?

sample-mpp-module/build.gradle
```diff
@@ -8,20 +8,20 @@ android {
 }
 
 kotlin {
-    def configureFrameworks = { target ->
-        target.binaries {
-            framework {
-                baseName = "shared"
-                export project(':reaktive')
+    if (System.getenv().keySet().contains("XCODE_VERSION_MAJOR")) {
```
CherryPerry, line 11: Add a comment here that envvar is available only when the task is run from XCode as pre-built step script.

sample-android-app/build.gradle
arkivanov: I don't mind, updated. 33 was the minimum required.

sample-mpp-module/build.gradle
arkivanov: Done.

CherryPerry (approved):

